PR #3, opened 2025-03-07
FEAT: Generate ODS Fact Tables

rymarczy:
This change contains the initial business logic for generating ODS Fact tables for ODS Qlik History datasets. 

This business logic allows for incremental loading of ODS Fact table from History datasets sources. Progress is tracked by `header__change_seq` column that is available from Qlik CDC records. This process adds an `odin_index` field to all Fact tables so that records can be referenced by a single index column. 

This change required the addition of a couple new Parquet utility functions. 

- fast_last_mod_ds_max
  -  faster way to get the maximum value from a dataset column, if you know the maximum value will be in the last modified file of a parquet dataset
- ds_limit_k_sorted
  - Produces a limited number of sorted results from a very large parquet dataset in a memory constrained manner
- ds_batched_join
  - Join dataframe records to a very large parquet dataset in a memory constrained manner

There are some areas where the Fact table business logic can be streamed lined, however this initial version of the logic is functional, if not pretty.


src/odin/ingestion/qlik/cubic_archive.py
```diff
@@ -90,15 +87,7 @@ def __init__(self, table: str) -> None:
...
         self.export_folder = os.path.join(DATA_SPRINGBOARD, CUBIC_QLIK_DATA, self.table)
-        self.reset_tmpdir(make_new=True)
-
-    def reset_tmpdir(self, make_new: bool = True) -> None:
-        """Reset TemporaryDirectory folder."""
-        if hasattr(self, "_tdir"):
-            self._tdir.cleanup()  # type: ignore
-        if make_new:
-            self._tdir = tempfile.TemporaryDirectory()
-            self.tmpdir = self._tdir.name
```
rymarczy, line 101: Moved this to OdinJob to make the tempdir functionality available to all OdinJob objects be default.

src/odin/generate/cubic/ods_fact.py
```diff
@@ -0,0 +1,372 @@
...
+        self.part_columns = []
+        if "edw_inserted_dtm" in self.history_ds.schema.names:
+            self.part_columns.append("odin_year")
+
+        history_ds_rows = self.history_ds.count_rows()
+        history_ds_groups = 0
+        frag: pd.ParquetFileFragment
+        for frag in self.history_ds.get_fragments():
+            history_ds_groups += frag.num_row_groups
+        self.batch_size = int(history_ds_rows / (20 * history_ds_groups))
```
huangh, line 136: I'm used to extracting out "magic numbers" in my code - would it be possible/make sense to bubble these values up into a config struct, defined at the top of the file maybe?

src/odin/generate/cubic/ods_fact.py
```diff
@@ -0,0 +1,372 @@
...
+                    download_object(found_objs[-1].path, destination)
+                    sync_paths.append(destination)
+        else:
+            part_columns = None
+            found_objs = list_objects(f"{self.s3_export}/", in_filter=".parquet")
+            if found_objs:
+                sync_file = found_objs[-1].path.replace("s3://", "")
+                destination = os.path.join(self.tmpdir, sync_file.replace("/year_", "/temp_"))
+                download_object(found_objs[-1].path, destination)
+                sync_paths.append(destination)
```
huangh, line 166: This if/else looks to be used to populate search_path. Can we generate search_paths in a list (or just the default path if part_column = None), then list comprehension to generate sync_paths? wrap found_objs...download in a method download_found_objects(...) to be the lambda in list comprehension. 

This deduplicates code, and minimizes the amount of code in the control flow, which is more clear imo.

src/odin/generate/cubic/ods_fact.py
```diff
@@ -0,0 +1,372 @@
...
+            )
+            if "edw_inserted_dtm" in write_schema.names:
+                batch = batch.append_column(
+                    "odin_year",
+                    pa.array(pc.strftime(batch.column("edw_inserted_dtm"), "%Y"), type=pa.int32()),
+                )
+            batch = batch.drop_columns(self.history_drop_columns)
+            writer.write_batch(batch)
+            odin_index += batch.num_rows
+            if os.path.getsize(write_path) * 3 > free_disk_bytes():
```
huangh, line 221: bubble up this tunable parameter?

src/odin/generate/cubic/ods_fact.py
```diff
@@ -0,0 +1,372 @@
...
+        writer.close()
+        self.sync_tmp_paths([write_path])
+        self.reset_tmpdir()
+
+    def load_cdc_records(self) -> int:
+        """Load change records from history files."""
+        fact_ds = ds_from_path(f"s3://{self.s3_export}/")
+        _, max_fact_seq = ds_column_min_max(fact_ds, "header__change_seq")
+        cdc_filter = (pc.field("header__change_seq") > max_fact_seq) & (
+            pc.field("header__change_oper") != "B"
```
huangh, line 239: can we explain this "B" in a comment?

src/odin/generate/cubic/ods_fact.py
```diff
@@ -0,0 +1,372 @@
...
+            partition_columns=part_columns,
+            export_folder=os.path.join(self.tmpdir, self.s3_export),
+            export_file_prefix="year",
+        )
+
+        for new_path in new_paths:
+            move_path = new_path.replace(f"{self.tmpdir}/", "")
+            upload_file(new_path, move_path)
+
+        if cdc_iter > 1:
```
huangh, line 333: Not sure what this means

huangh: I see. can we set a variable and name it something like rerun_quick = 60*5, rerun_next_hour = 60\*60, and return those instead?

src/odin/job.py
```diff
@@ -46,6 +52,8 @@ def start(self, schedule: sched.scheduler) -> None:
             )
 
         except Exception as exception:
+            # Don't run again for 12 hours on failure
+            run_delay_secs = 60 * 60 * 12
```
huangh, line 56: 👌

src/odin/utils/parquet.py
```diff
@@ -159,6 +167,139 @@ def ds_column_min_max(
     return (r["min"], r["max"])  # type: ignore[assignment,index]
 
 
+def fast_last_mod_ds_max(partition: str, column: str) -> Any:
+    """
+    Find max value of column from the file of parquet partition that was most recently modified.
+
+    This is usefull for very large datasets where it is guaranteed that the max value of a column
```
huangh, line 174: usefull->useful

src/odin/utils/parquet.py
```diff
@@ -159,6 +167,139 @@ def ds_column_min_max(
...
+    will be in the most recently modified file of the partition.
+
+    TODO: make this work locally as well??
```
huangh, line 177: yes please 😄

src/odin/generate/cubic/ods_fact.py
```diff
@@ -0,0 +1,372 @@
...
+        self.snapshot_source = f"{self.s3_source}/snapshot={self.history_snapshot}/"
+        self.history_ds = ds_from_path(f"s3://{self.snapshot_source}")
+
+        self.part_columns = []
+        if "edw_inserted_dtm" in self.history_ds.schema.names:
+            self.part_columns.append("odin_year")
+
+        history_ds_rows = self.history_ds.count_rows()
+        history_ds_groups = 0
+        frag: pd.ParquetFileFragment
```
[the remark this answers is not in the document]
rymarczy: Yup, just for linting. It's one of the ways to type hint for loop iter variables.

src/odin/generate/cubic/ods_fact.py
rymarczy: Yeah, I totally agree. I want to make a custom `dataset` class that offers a bunch of methods and properties like this automatically.

src/odin/generate/cubic/ods_fact.py
rymarczy: I don't think this would be particularly "tunable". Just trying to make sure the parquet files doesn't exceed 1/4 of available disk space.

requirements.txt
huangh, line 1: are we able to do anything actionable were these hashes to no longer match? or would we just grab the package regardless...because it's a dependency we need? 

If there's no process for actively resolving mismatched hash, i would be inclined to remove this all to simplify the requirements.txt

pyproject.toml
```diff
@@ -15,12 +15,12 @@ dependencies = [
     "boto3",
     "polars",
     "psutil",
-    "pyarrow>=19.0.0",
+    "pyarrow",
```
huangh, line 18: related to the comment in requirements.txt...these are unpinned, so the requirements.txt hashes will keep changing/lots of commit churn.

huangh (approved):
if there's a reason for the hashes, then i'm good. otherwise i think removing is a good idea

src/odin/generate/cubic/ods_fact.py
huangh: 👌🏼

src/odin/generate/cubic/ods_fact.py
huangh: im ok revisiting this discussion if we ever move common functionality out of our individual products

requirements.txt
rymarczy: The `requirements.txt` files are currently created by the `config_venv.sh` script. 

That script won't update any packages unless a package in a requirements.txt file specifically doesn't agree with a package version defined in pyproject.toml. So with no versions specific in pyproject.toml, the versions/hashes defined in the requirements.txt files just happen to agree with the latest available when I last ran `config_venv.sh` to generate new requirements.txt files. 

The hashes are more so to confirm that the packages pulled from a package repository have not been corrupted or replaced by some man in the middle type attack. 

The `config_venv.sh` script could be upgraded to allow for upgrading of all, or specific, packages to @latest with some additional arguments. However, this could also be handled by updating pyproject.toml to specify >= package versions and forcing package updates when running `config_venv.sh` as it currently exists.

pyproject.toml
rymarczy: It's not a perfect system, but there won't be any regular churn with the way `config_venv.sh` is currently configured to run.

PR #5, opened 2025-03-28
FIX: run each Job in Process to free resources

rymarczy:
Constructing jobs as python Class objects seems to cause an issue with clearing resources between job runs. I believe this is related to the long-lived classes hanging on to resources after a job has completed. 

To resolve this issue, this change runs each OdinJob inside of a process using the `job_proc_schedule` function as the `sched.scheduler` callback. 

This change was deployed to Dev yesterday evening and resulted the the following change to memory usage:
image

As can be seen, memory usage between job runs now drops to almost 0.

### DIFF
--- a/src/odin/generate/cubic/ods_fact.py
+++ b/src/odin/generate/cubic/ods_fact.py
@@ -1,4 +1,5 @@
 import os
+import sched
 from typing import List
 from typing import Tuple
 
@@ -9,6 +10,7 @@ import pyarrow.parquet as pq
 import pyarrow.dataset as pd
 
 from odin.job import OdinJob
+from odin.job import job_proc_schedule
 from odin.utils.logger import ProcessLog
 from odin.utils.logger import free_disk_bytes
 from odin.utils.locations import DATA_SPRINGBOARD
@@ -31,6 +33,7 @@ from odin.utils.aws.s3 import download_object
 from odin.utils.aws.s3 import upload_file
 from odin.ingestion.qlik.dfm import dfm_from_s3
 from odin.ingestion.qlik.dfm import QlikDFM
+from odin.ingestion.qlik.tables import CUBIC_ODS_TABLES
 
 NEXT_RUN_DEFAULT = 60 * 60  # 1 hour
 NEXT_RUN_IMMEDIATE = 60 * 5  # 5 minutes
@@ -436,3 +439,14 @@ class CubicODSFact(OdinJob):
             return NEXT_RUN_IMMEDIATE
 
         return next_run_secs
+
+
+def cubic_ods_gen_fact_schedule(schedule: sched.scheduler) -> None:
+    """
+    Schedule All Jobs for generate cubic ODS fact tables process.
+
+    :param schedule: application scheduler
+    """
+    for table in CUBIC_ODS_TABLES:
+        job = CubicODSFact(table)
+        schedule.enter(0, 1, job_proc_schedule, (job, schedule))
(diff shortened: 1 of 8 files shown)


src/odin/utils/parquet.py
```diff
@@ -17,8 +16,7 @@
 from itertools import chain
 
 import polars as pl
-from polars.type_aliases import ColumnNameOrSelector
-from polars.type_aliases import PolarsDataType
+from polars._typing import SchemaDict
```
rymarczy, line 19: updated Polars package was throwing warnings to stop using `polars.type_aliases` This resulted in having to change some other typing stuff for this file..

src/odin/run.py
```diff
@@ -50,26 +50,16 @@ def start():
...
-        job = ArchiveCubicQlikTable(table)
-        scheduler.enter(0, 1, job.start, (scheduler,))
-        job = CubicODSFact(table)
-        scheduler.enter(0, 1, job.start, (scheduler,))
-
-    scheduler.run()
+    # Schedule ODIN Jobs
+    schedule_sigterm_check(schedule)
+    schedule_cubic_archive_qlik(schedule)
+    schedule_cubic_ods_fact_gen(schedule)
```
rymarczy, line 60: Now using single function to schedule each job type for application

src/odin/generate/cubic/ods_fact.py
```diff
@@ -436,3 +439,14 @@ def run(self) -> int:
             return NEXT_RUN_IMMEDIATE
 
         return next_run_secs
+
+
+def schedule_cubic_ods_fact_gen(schedule: sched.scheduler) -> None:
```
rymarczy, line 444: move scheduling to job

src/odin/ingestion/qlik/cubic_archive.py
```diff
@@ -340,3 +344,22 @@ def run(self) -> int:
             next_run_secs = 60 * 60
 
         return next_run_secs
+
+
+def schedule_cubic_archive_qlik(schedule: sched.scheduler) -> None:
```
rymarczy, line 349: move scheduling to job

src/odin/job.py
```diff
@@ -63,4 +65,20 @@ def start(self, schedule: sched.scheduler) -> None:
 
         finally:
             self.reset_tmpdir()
-            schedule.enter(run_delay_secs, 1, self.start, (schedule,))
+
+        return run_delay_secs
+
+
+def job_proc_schedule(job: OdinJob, schedule: sched.scheduler) -> None:
```
rymarczy, line 72: Run all jobs in separate process. When process is finished all job resources are released.

src/odin/job.py
```diff
@@ -63,4 +65,20 @@ def start(self, schedule: sched.scheduler) -> None:
...
+    """
+    Odin Job Runner as Process.
+
+    This function runs each OdinJob as it's own process so resource usage can be fully cleared
+    between job runs.
+
+    :param job: Job to be run and re-scheduled
+    :param schedule: main application scheduler
+    """
+    with get_context("spawn").Pool(1) as pool:
```
huangh, line 82: this works how you want without the maxchildtasks=1 thing?

huangh (approved):
the proof is in the dev. 🧁

src/odin/job.py
rymarczy: Yup, since it's only running one thing at a time it's essentially the same as `maxchildtasks=1` if you're running multiple processes.

PR #6, opened 2025-03-31
FIX: Stop Qlik Archive process on CDC process Error

rymarczy:
Previously, when an error was encountered with processing a Qlik CDC file, the ArchiveCubicQlikTable process would log the error and move the CDC file to an error location for potentially future re-processing. 

Odin does not have a process for re-processing CDC error files and the ArchiveCubicQlikTable and CubicODSFact processes rely on loading qlik change records in contiguous order.

This change modifies the ArchiveCubicQlikTable to raise and stop when encountering an issue with a CDC file. This will stop any future processing for that table and allow time for a fix to be implemented.


huangh (approved):
🧠

PR #7, opened 2025-03-31
FIX: Coalesce NULL edw_inserted_dtm values

rymarczy:
The CubicODSFact table process relies on partitioning by the year of the `edw_inserted_dtm` column. If his column is null, the process currently fails. (this happened for EDW.FARE_PROD_USERS_LIST_DIMENSION)

This change updates the `edw_inserted_dtm` partition process to coalesce a Null with 0 so that they `odin_year` value can also be used as an int32.


src/odin/generate/cubic/ods_fact.py
```diff
@@ -257,7 +257,9 @@ def load_new_snapshot(self) -> None:
             if "edw_inserted_dtm" in write_schema.names:
                 batch = batch.append_column(
                     "odin_year",
-                    pa.array(pc.strftime(batch.column("edw_inserted_dtm"), "%Y"), type=pa.int32()),
+                    pc.coalesce(pc.strftime(batch.column("edw_inserted_dtm"), "%Y"), "0").cast(
+                        pa.int32()
+                    ),
```
huangh, line 262: the default is int64, right? this might not matter if the field is literally just a year, but i think this introduces a bit more computation and could introduce errors if edw_inserted_dtm is every a real timestamp.

okay if this is moot

huangh (approved):

src/odin/generate/cubic/ods_fact.py
rymarczy: The default of the `strftime`, for the `odin_year` creation, is `string`. 

Odin year gets turned into a parquet partition so `odin_year=YYYY` on the parquet file path. When reading the default usually depends on the library that's reading the parquet file. For pyarrow it's `int32`.

`edw_inserted_dtm` is almost always a real timestamp. It's only not a timestamp for 1 record of one table on Qlik PROD data. 

Is that the default you're talking about?

PR #8, opened 2025-04-02
[FEAT] New parquet dataset min_max function based on dataset metadata

rymarczy:
The `pyarrow.acero` based `ds_column_min_max` function was becoming slow on very large PROD ODS datasets. 

This change adds a parquet dataset min_max function based on the dataset metadata. It only pulls file metadata so is very fast for very large parquet datasets. This will only work if the dataset has statistics written to the parquet file, which all ODIN parquet files do. 

For the `cubic_qlik/EDW.DEVICE_EVENT` dataset, which is about 10GB in size (~319,810,572 rows):

```
%timeit -n 5 -r 5  ds_metadata_min_max(ds, "header__change_seq")
# 492 ms ± 254 ms per loop (mean ± std. dev. of 5 runs, 5 loops each)
# mem usage ~ 180mb

%timeit -n 1 -r 1  ds_column_min_max(ds, "header__change_seq")
# 1min 55s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
# mem usage ~640mb
```

### DIFF
--- a/src/odin/generate/cubic/ods_fact.py
+++ b/src/odin/generate/cubic/ods_fact.py
@@ -13,13 +13,14 @@ from odin.job import OdinJob
 from odin.job import job_proc_schedule
 from odin.utils.logger import ProcessLog
 from odin.utils.logger import free_disk_bytes
+from odin.utils.runtime import sigterm_check
 from odin.utils.locations import DATA_SPRINGBOARD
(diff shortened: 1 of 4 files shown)


huangh (approved):
✌🏼

PR #9, opened 2025-04-04
FEAT: DEV Pipeline for S&B AFC API Archive Job

rymarczy:
This is an initial implementation for the S&B AFC API Archive job. 

The API, as it currently stands, has no functional pagination method, so this should not be deployed to PROD. 

The API works by providing csv.gz files as downloadable blobs. This Job downloads those blobs, converts them to parquet and then merges the parquet with any existing parquet files in S3. 

This implementation also does not do any file path partitioning. Not totally sure if this will or will not be needed in the future. 

This is in conjunction with a devops PR: https://github.com/mbta/devops/pull/2745 that made the required secrets available to the ECS in AWS.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -5,3 +5,7 @@ DATA_ARCHIVE=mbta-ctd-dataplatform-dev-archive
 DATA_ERROR=mbta-ctd-dataplatform-dev-error
 DATA_INCOMING=mbta-ctd-dataplatform-dev-incoming
 DATA_SPRINGBOARD=mbta-ctd-dataplatform-dev-springboard
+
+# AFC API
+AFC_API_CLIENT_ID=secret
+AFC_API_CLIENT_SECRET=secret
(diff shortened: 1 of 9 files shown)


huangh (approved):
do you already have the cleaning up /tmp folder logic? the csvs getting pulled down won't blow the container ephemeral storage? Not sure if worth putting a check for that since you've stated clearly "don't deploy to prod" caveats, but otherwise looks good to me.

rymarczy:
> do you already have the cleaning up /tmp folder logic? the csvs getting pulled down won't blow the container ephemeral storage? Not sure if worth putting a check for that since you've stated clearly "don't deploy to prod" caveats, but otherwise looks good to me.

Reset of /tmp folders happens automatically for every `OdinJob` at task completion: https://github.com/mbta/odin/blob/821df630fd5345dc4cc8162f1015e4c51b0df46e/src/odin/job.py#L67

rymarczy:
devops PR to run job: https://github.com/mbta/devops/pull/2839

PR #10, opened 2025-04-07
remove unneeded configurability

skyqrose:
Asana Task: [Make Odin more configurable](https://app.asana.com/0/1209584755824899/1209886680141188/f)

I'm planning to refactor how config is specified. That will be easier if there are fewer things that need to be configured between different environments. So this removes configurability for a couple things that don't need it.

No changes in prod. Minor changes when developing locally.
- Always set the number of threads is 2x the number of cpus.
  - This was already the case in AWS.
  - This doubles the number of threads when not running in AWS.
- Remove timestamp from local logging.
  - It was already not used in AWS.
  - This is a change when not running locally, but the timestamp isn't really that useful anyway and kind of noisy.
- Remove the SERVICE_NAME env var
  - It was set in the app and only used in one place, so this removes some code with no change in behavior.

### DIFF
--- a/src/odin/utils/logger.py
+++ b/src/odin/utils/logger.py
@@ -22,10 +22,8 @@ def free_disk_bytes() -> int:
 MdValues = Optional[Union[str, int, float]]
 
 DATE_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
-LOG_FORMAT = "{asctime} {levelname:>8s} {message}"
-# if running on AWS drop timestamp because Splunk handles that
-if bool(os.getenv("AWS_DEFAULT_REGION")):
-    LOG_FORMAT = "{levelname:>8s} {message}"
+# skip "{asctime}" timestamp because Splunk handles that.
+LOG_FORMAT = "{levelname:>8s} {message}"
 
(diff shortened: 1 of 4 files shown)


src/odin/utils/logger.py
```diff
@@ -22,12 +22,10 @@ def free_disk_bytes() -> int:
 MdValues = Optional[Union[str, int, float]]
 
 DATE_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
-LOG_FORMAT = "{asctime} {levelname:>8s} {message}"
-# if running on AWS drop timestamp because Splunk handles that
-if bool(os.getenv("AWS_DEFAULT_REGION")):
-    LOG_FORMAT = "{levelname:>8s} {message}"
```
rymarczy, line 28: This check can stay in. It is needed to print timestamp information to the console when running the application not in AWS.

rymarczy (changes requested):
Just 1 request.

src/odin/utils/logger.py
skyqrose: k, reverted this change

rymarczy (approved):
:+1:

PR #12, opened 2025-04-11
FEAT: Run OdinJob in Process instead of Pool

rymarczy:
This change modified how OdinJob's are run, using a Multiprocessing Process instead of a Pool. 

The Pool construct would hang if a process, started by the Pool, was killed by the kernel because of something like an OOM error. The Process construct returns gracefully in this situation, providing a non-zero exitcode to indicate what happened to the child process. 

The Process construct does not offer simple return values like the Pool does. So this change also implements a process-safe shared Value object, to return `run_delay_secs` from the OdinJob.start function.

### DIFF
--- a/src/odin/job.py
+++ b/src/odin/job.py
@@ -76,6 +76,10 @@ def job_proc_schedule(job: OdinJob, schedule: sched.scheduler) -> None:
     This function runs each OdinJob as it's own process so resource usage can be fully cleared
     between job runs.
 
+    Running the job in a `Process` also allows the scheduler to be unaffected by jobs that are
+    killed by the machine kernel because of something like an OOM error. The exitcode of each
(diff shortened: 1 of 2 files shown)


grejdi-mbta (approved):
Where are we in capturing this in an ADR? Please add a task for it.

This is a side thought, but what do you think about using [Supervisor](https://supervisord.org) for managing processes? Feels like you are building some of the same functionality.

rymarczy:
> Where are we in capturing this in an ADR? Please add a task for it.
> 
> This is a side thought, but what do you think about using [Supervisor](https://supervisord.org) for managing processes? Feels like you are building some of the same functionality.

This is the first I'm hearing about Supervisor, I'll create an asana ticket to look into it's use.

PR #14, opened 2025-04-11
FEAT: New ds_metadata_limit_k_sorted function for parquet

rymarczy:
This change adds a new `ds_metadata_limit_k_sorted` function to pull records from large parquet datasets in a more performant manner in some situations.

### DIFF
--- a/src/odin/utils/parquet.py
+++ b/src/odin/utils/parquet.py
@@ -8,6 +8,7 @@ from typing import Optional
 from typing import List
 from typing import Any
 from typing import Literal
+from typing import TypedDict
 from functools import reduce
 from operator import gt
 from operator import lt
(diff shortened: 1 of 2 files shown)


src/odin/utils/parquet.py
```diff
@@ -319,10 +374,125 @@ def ds_limit_k_sorted(
...
 
-    log.complete(num_results=result_batch.num_rows)
     return_df = pl.from_arrow(result_batch)
     if isinstance(return_df, pl.Series):
         raise TypeError("Always dataframe.")
+    log.complete(num_results=result_batch.num_rows)
+    return return_df
+
+
+def ds_metadata_limit_k_sorted(
```
huangh, line 384: (suggestion/question) limit k sorted implies that I can set a "k" and get up to "k" sorted items, but I can only set a max_nbytes to hope to get the k'th element here right? if so, perhaps a different name for the method to make it clear its by bytes.

not sure if this is an internal-only method or used to get results for end users, but I assume most end user queries would want something like the first "k" elements matching a filter.

tests/utils/parquet_test.py
```diff
@@ -162,3 +170,66 @@ def test__pq_find_part_offset(tmp_path) -> None:
 
     os.remove(file2)
     assert _pq_find_part_offset(tmp_fldr) == 2
+
+
+def test_pq_path_partitions(pq_files) -> None:
```
huangh, line 175: maybe worth adding something here to test max memory usage given large parquet sets?

src/odin/utils/parquet.py
rymarczy: I added an optional `max_rows` parameter to limit to a specific set of rows. 

In general this is a function meant to constrain memory usage so the `max_nbytes` parameter will be more widely used.

tests/utils/parquet_test.py
rymarczy: I don't have any current scaffolding to text max memory usage, and the function makes no guarantees to limit to a specific amount of memory usage. I would say the implied guarantee  is that it will produce consistent memory usage, but the specific amount could vary widely depending on application.

huangh: sounds good for now. I was more thinking that if someone made a change in the future that broke this implicit guarantee, then we'd want to catch it somehow. but i get that its hard to pin down in a test.

PR #15, opened 2025-04-14
FEAT: Super Simple Migration Process

rymarczy:
This change gives odin the ability to run super simple one-way migrations. 

This will generally be applied to needing to do on-time move or modify operations on S3 objects. 

This process is explained by the `start_migrations()` function description:
```
    This process allows ODIN to run one-time migrations based on S3 "status_files" being used as a
    checkpoint to determine which migrations have been run and/or need to be run.

    To create/run a new migration, create a new incremented python file in the folder of the
    ECS task/environment you want to run a migration for.

    For migration on odin dev environment (task_name=odin-dev):
       -> odin > migrate > migrations > odin-dev > 000X.py
    Note:
       Any files in the migration folder not following the migration file format will
       be skipped/ignored.

    Inside of the 000X.py file, create a `migration()` function and accepts and returns nothing.
    This is the function that will be called during the migration process.

    If any exception is thrown during the migration process, the migration will be considered
    to be "failed" and the ECS will enter an infinte wait state untill the failure can be resolved.

    If the migration complets successfully, a file will uploaded to S3 with the migration number
    `000X`. Future migration attempts will skip this migration as all migrations are run in
    incremental order.
```


grejdi-mbta (commented):
I think ADR would be helpful for this, with a clear example of some S3 calls made. You can use the `dataplatform-local` bucket to work something in.

src/odin/run.py
```diff
@@ -47,6 +48,7 @@ def start():
             "ECS_TASK_GROUP",
         ],
     )
+    start_migrations()
```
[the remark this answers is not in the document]
rymarczy: I've moved `start_log_file` to another location to avoid confusion, but that is just a function to be used for local development. 

`run.py:start()` is always the primary entry point for `odin` as an application. 

The purpose of the `start()` function is currently the following:
1. validate environment is set with `validate_env_vars`
2. Run any migrations for the task `start_migrations`
3. Schedule all ODIN job types that will be running
4. Start the scheduler.

src/odin/run.py
```diff
@@ -47,40 +40,15 @@ def start():
...
-    log = ProcessLog("odin_event_loop")
+    ProcessLog("odin_event_loop")
 
     # Schedule ODIN Jobs
     schedule_sigterm_check(schedule)
     schedule_cubic_archive_qlik(schedule)
     schedule_cubic_ods_fact_gen(schedule)
 
     schedule.run()
-    log.complete()
```
rymarczy, line 61: Dropped because this will never actually get called as the ECS is always in the `schedule` loop.

src/odin/run.py
```diff
@@ -47,40 +40,15 @@ def start():
...
-
-
-def start_log_file():
```
rymarczy, line 64: Moved to `/utils/run_log_to_file.py`

src/odin/migrate/process.py
```diff
@@ -0,0 +1,128 @@
...
+    :param stem: 4 numbers of migration stem. (0001)
+    """
+    with tempfile.TemporaryDirectory() as tmpdir:
+        file_path = os.path.join(tmpdir, stem)
+        with open(file_path, "w") as f:
+            f.write(stem)
+        upload_file(file_path, os.path.join(status_path, stem))
+
+
+def run_task_migration(modules_path: str, task_name: str) -> None:
```
rymarczy, line 44: renamed this to be more explicit with purpose

rymarczy:
@grejdi-mbta ADR draft added to PR.

docs/adr/migrations/README.md
```diff
@@ -0,0 +1,138 @@
...
+Not easily reversible. May be impossible to test migrations before deployment.
+
+
+## Notes
+
+
+Example migration `/migrations/odin-dev/0001.py`:
+
+```py
+def migration() -> None:
```
grejdi-mbta, line 133: I was thinking a more true to nature migration, like list objects under an S3 prefix. The ADR looks good for me. Thanks for adding.

grejdi-mbta (commented):
I'll let @huangh provide the 👍 .

docs/adr/migrations/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+
+### Related artifacts
+
+Each odin process task-name will have a folder in [migrations](../../../src/odin/migrate/migrations).
+
+
+### Related principles
+
+Not easily reversible. May be impossible to test migrations before deployment.
```
huangh, line 124: What do we mean by "impossible to test migrations"? This scares me. 

Is there a possibility of this being implemented where we can have a "dry run" to test without overwriting? Or a version where we can run locally, or operate on a subset to verify before running the whole migration? 

We're migrating S3 files, so only delete/create operations available. Could we do something where we copy to a temp object (S3_object.original), perform the migration, (S3_object delete; S3_object create w/ migration)? This would temporarily double the size of the data on S3, but this way we could verify the migration across the whole set, or retain the temporary for a time until we are sure the migration was good.

src/odin/migrate/process.py
```diff
@@ -0,0 +1,128 @@
...
+
+    Note:
+        Any files in the migration folder not following the migration file format will
+        be skipped/ignored.
+
+    Inside of the 000X.py file, create a `migration()` function and accepts and returns nothing.
+    This is the function that will be called during the migration process.
+
+    If any exception is thrown during the migration process, the migration will be considered
+    to be "failed" and the ECS will enter an infinite wait state until the failure can be resolved.
```
huangh, line 103: What is the intended way to resolve this error case if the migration is "failed"? We can't access the infinitely spinning ECS task bc we don't have permissions either way, right?

src/odin/migrate/process.py
```diff
@@ -0,0 +1,128 @@
...
+        except Exception as exception:
+            log.failed(exception)
+            raise exception
+
+
+def start_migrations():
+    """
+    Start Simple Migration Process.
+
+    This process allows ODIN to run one-time migrations based on S3 "status_files" being used as a
```
huangh, line 86: Would there ever be a case where some new inputs require the full set of migrations, but the existing inputs have already been migrated to "000X"?

huangh: More concretely, if there is a schema change or something, wouldn't live incoming data (if it exists) need to be converted to the new schema simultaneously? Is this a concern?

docs/adr/migrations/README.md
huangh: I can understand the rationale for having the catch-all migrate() method to allow for flexibility in unknown migrations to occur in the future, but perhaps splitting this up into the potential operations (limited in scope bc of S3) would make it clear that a certain migration is "COPY/UPDATE only" or "DELETE only" etc. 

e.g. 

migrate_add() 
migrate_delete()
migrate_modify() // this would call migrate_add() to generate a S3_obj.orig via migrate_add(), then upload the modified version via a second migrate_add()

migrate_validate() // optional step to validate state of changes, if desired, and could call migrate_delete() to clear out temps if all validated. 

migrate() would call a series of these - they would all have structured logging

src/odin/migrate/process.py
rymarczy: I do not imagine such a scenario would occur, as migrations are meant to be environment specific one time actions. 

I'm not sure I understand your scenario. A schema change, as indicated by a migration, would be equivalent to a restart of the most up to date version of a job's operation.

docs/adr/migrations/README.md
rymarczy: I'm not aware of way to implement "dry run" boto3 S3 operations. It would require the mocking of situation and environment specific S3 scenarios, in which case the development of the mocking would be as prone to error as the development of any individual migration action. 

The potential size of datasets that migrations could impact makes copying entire datasets for some local mocking process, while parallel versions of these jobs are still running, subject to race conditions and additional complexity. 

Personally, I don't like the idea of per-defining migration functions as their utility would be entirely negated as soon as a process didn't fall into one of the per-defined action buckets.  I would also prefer not to hide the complexity of migrations behind any abstractions. Each migration should created with explicit and straight-forward business logic. 

This framework puts pressure on the developer to create well defined and well documented migration operations. I consider this a positive aspect of the model, as any helper functions or dry-run frameworks could provide a false sense of security regarding the success of a migration action.

src/odin/migrate/process.py
huangh: 👌

src/odin/migrate/process.py
huangh: 👌

docs/adr/migrations/README.md
huangh: I think dry run might be too loaded a term, but I'm still not sure we need to commit to a destructive migration from the get go, when we could instead write "new" temporary files with the migrated characteristics, and compare before applying another migration to delete the old and replace with the new. Would it not be safer to enforce this - in code - as a two-stage process?

Perhaps there are other migrations that I'm not considering, but if the migrations are "S3 object at rest -> S3 object at rest", then the only operations are S3 operations, which are limited in scope was my only point.

rymarczy: I think I understand the idea, but I don't really see how it would be implemented in practice.

1. Would this 2 step process be meant to run on different environments or the same environment? 
2. Is this 1st step meant to be run locally or on an ECS?
3. This would involve doing a migration, before the actual migration?
4. How would the stages work, would the confirmation be automated? Or would you run the first stage, perform some manual confirmation and then run the second stage? 

I agree that the concept here is simple, but the execution adds a layer of complexity that doesn't guaranteed a successful outcome. We could be talking about hundreds of thousands of S3 objects needing to be moved or modifying datasets that are 100's of GBs or TBs in size. 

Something like this may work on small datasets, but as soon as you reach complexity of scale I can't see this working very well. We've had similar issues in the past with database migrations, where we run testing locally or on dev environments that are successful, but that execution at scale causes issues that are impossible to replicate outside of a PROD environment. 

The best way to handle this would be to have a staging environment that is a complete mirror of PROD. In that case, migrations could be testing on staging and then moved to prod if successful.

huangh: 1) 2 step process is meant to be run on the same environment
2) In ECS
3) Yes
4) Could have a method that automatically validates and confirms, or could manually confirm and then run second stage. 

The idea i keep coming back to is that AWS isn't really modifying an object in place - it is replacing it. And parquet files aren't mutable either. So no matter what, your migration will involve making and writing a new object to S3. What if that S3 object were tagged with the migration version suffix? Or it's in a new partition? Then, we write a new one, and delete the old one (after some validating). 

I think this is similar to the staging == prod idea, but doing it in this migration job ensures that it is 100% a mirror, where as staging == prod may not be 100% the case (depending on downtime, last deploy, etc)

huangh: Agreed, experience and time can guide future work.

huangh (approved):

PR #16, opened 2025-04-16
chore: Drop INCOMING bucket env var

rymarczy:
Incoming bucket is not used in any ODIN operations, can be dropped as env var. 

This change will require a devops PR before merging. 

Devops PR blocked by terraform_modules PR: https://github.com/mbta/terraform_modules/pull/163

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -3,5 +3,4 @@
 # S3 Buckets
 DATA_ARCHIVE=mbta-ctd-dataplatform-dev-archive
 DATA_ERROR=mbta-ctd-dataplatform-dev-error
-DATA_INCOMING=mbta-ctd-dataplatform-dev-incoming
 DATA_SPRINGBOARD=mbta-ctd-dataplatform-dev-springboard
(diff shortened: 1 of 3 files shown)


grejdi-mbta (commented):
I don't think you should drop this. It reflects the infrastructure that you are working in, even though it's not being used. That coupled with a coordination with a DevOps PR, it feels like too much for not that much gain. One other thing, @skyqrose  might need the Incoming bucket as that's where the RF good list is stored right now.

skyqrose (approved):
I think dropping it is fine. I'm hoping to change things to configure each job more independently. The incoming location for one job might be different than the incoming location for another. So if I do use it, it would end up in the job configuration rather than a global env var.

(Side note: The RF list probably won't be processed via Odin, so that specific use case doesn't apply here, but I do expect to use the incoming bucket for other paratransit data processing.)

rymarczy:
I agree with @grejdi-mbta here. It reflects the infrastructure of the data-platform buckets.

PR #17, opened 2025-04-25
FIX: Odin Migration to reset ODS Process

rymarczy:
During initial launch of ODIN Cubic ODS processing, the process ran into several errors that tainted the state of the Archive and Fact tables produced by the process.

This change slightly modified the archive and fact table processes, to avoid process failures and also executes a migration to reset the state of S3 objects for the process. 

The main changes to the Cubic ODS processing are:
  1. Do not make up `header__change_seq` values for LOAD records.
  2. Store `header__change_seq` as string instead of Decimal type
  3. Properly sort CDC files from all import locations so that files are processed in order based on timestamp file name. 

The Migration process is described in the migration files. 

New S3 permissions required to execute migration are added here: https://github.com/mbta/terraform_modules/pull/199


src/odin/ingestion/qlik/cubic_archive.py
```diff
@@ -364,13 +363,5 @@ def schedule_cubic_archive_qlik(schedule: sched.scheduler) -> None:
     :param schedule: application scheduler
     """
     for table in CUBIC_ODS_TABLES:
-        # This will be not be a permanent part of the pipeline and can be removed in the future
```
rymarczy, line 367: This has been executed and does not need to be called anymore.

pyproject.toml
```diff
@@ -39,6 +39,9 @@ log_level = 20
 [tool.mypy]
 ignore_missing_imports = true
 python_version = 3.13
+exclude = [
```
rymarczy, line 42: mypy does not like the format of the migrations folder.

src/odin/migrate/migrations/odin-dev/0001.py
```diff
@@ -0,0 +1,49 @@
...
+from odin.utils.aws.s3 import rename_objects
+from odin.utils.locations import DATA_ARCHIVE
+from odin.utils.locations import DATA_SPRINGBOARD
+from odin.utils.locations import CUBIC_QLIK_DATA
+from odin.utils.locations import CUBIC_ODS_FACT_DATA
+from odin.utils.locations import CUBIC_QLIK_PROCESSED
+from odin.utils.locations import IN_QLIK_PREFIX
+
+
+def migration() -> None:
```
huangh, line 15: (suggestion/question) - I can't tell if it's a better practice to have the entire migration logic duplicated in each sub folder `odin-dev` vs `odin-prod` - or if it would be better to have an `odin-common/0001.py` that has the same migration code for both. 

i could see eventually something in the migration being different for prod/dev and not getting caught, which wouldn't happen if it were the same file.

huangh (approved):
🧁

src/odin/migrate/migrations/odin-dev/0001.py
rymarczy: I prefer to have them be explicitly separate. 

I think the benefit of having a common migration mechanism is negated by the complication of implementing a process that handles common and explicitly separate migration files.

PR #19, opened 2025-04-30
FIX: ODSFact Next Runn Immediate Logic

rymarczy:
This change updates the way we determine if an immediate next run is needed for the CubicODSFact job. 

An additional `ds_metadata_limit_k_sorted` call is made and if it returns a number of records that is close to the number of records being processed by the current event loop, then the process will run again right away.

This call is a little expensive, but it is worth it if a 4 hour delay in processing available data is avoided. 

The previous method was not running tables right away in expected situations.


huangh (approved):
🐤

PR #20, opened 2025-05-02
feat(spare): process Spare vehicles data into Parquet

skyqrose:
Asana Task: [Process Spare Dev data into Parquet with Odin](https://app.asana.com/1/15492006741476/project/1209584755824899/task/1209698688134966?focus=true)

This creates a new job which process data from Spare. For now, it's just a single API call (vehicles), but the plan is to add many more once the whole pipeline works end to end for vehicles.

Things I'm particularly unsure about and would like some guidance on:
- [x] What logging should this job do? I want to make sure any logging I do fits in with the rest of the app, has the right metadata, logs at the right times but not too noisily, etc. I wasn't sure, so right now it logs nothing (except the default from `job.py`).
- [x] Am I doing Python and Polars right? This is the first industrial-strength Python I've written in a few years, so I may be missing some things.
- [x] Are errors handled correctly?
  - Pretty much any error will cause a crash, and skip all the writes. And if there is a crash during the final writes, causing duplicate processing of some data, it's idempotent and won't cause any data corruption.
  - If it crashes due to a temporary issue like a network blip, it will get reprocessed on the next iteration, which is fine.
  - **I haven't made an error bucket**, so if it crashes due to a permanent issue (a bug, or unexpected shape of input data, or a corrupted output file), it will crash repeatedly on every run until a manual intervention. Is this okay? Even with an error bucket, with the types of errors I expect, it wouldn't be able to recover anyway and would still crash every time it got new input (though that would be much less frequently). Should I add an error bucket to reduce the noisiness of errors even if it won't make them recoverable?

PR in devops to turn the job on in dev: https://github.com/mbta/devops/pull/2845 (This job won't be turned on in prod for a while.) (The two PRs can go in either order.)


src/odin/spare/spare_job.py
```diff
@@ -0,0 +1,383 @@
...
+        self.api_endpoint = spare_api_endpoint
+        self.start_kwargs = {
+            "name": spare_api_endpoint["name"],
+            "api_path": spare_api_endpoint["api_path"],
+        }
+
+    def run(self) -> int:
+        """Entrypoint to run the job."""
+        run(self.config, self.api_endpoint, self.tmpdir)
+        next_run_secs = 1 * 60  # 1min
```
skyqrose, line 369: I set the frequency pretty high, once per minute, because consumers of this data will care about freshness. Most runs won't do much aside from check whether there's new input (there won't be).

Is this ok?

tests/spare/spare_job_test.py
```diff
@@ -0,0 +1,390 @@
...
+def test_load_parquet_local_not_exists(tmpdir):
+    path = os.path.join(tmpdir, "test.parquet")
+    df = spare_job.load_parquet(path, api_endpoint["columns"])
+    assert_frame_equal(df, empty_df)
+
+
+# datetime_from_delta_path_ms
+
+
+def test_datetime_from_delta_path_ms():
```
skyqrose, line 110: CI is failing cuz some of my tests don't have a docstring. I don't think a docstring would add much that's not covered in the test name or easy to see from the implementation, (in this case, the docstring would be `"""It works."""`) so I'd rather skip them. Is it possible to turn off the required docstring check on test files?

src/odin/spare/spare_job.py
```diff
@@ -0,0 +1,383 @@
...
+Note: Vehicle data changes rarely, and this file is optimized for that one case.
+As other endpoints with other needs are added, different logic will be added.
+
+Input files must have a timestamp in them.
+
+The output parquet file is logically append-only.
+Changes and deletions are appended to the end of the parquet file.
+Unchanged data is ignored, so that the parquet file does not grow unnecessarily large.
+
+Changes are based on the `id` field that Spare includes for every item in its API.
```
rymarczy, line 21: Based on this explanation, I would suggest modifying the structure of your parquet file. 

Generally columns created by odin, that do not exist in the original dataset are prefixed with `odin_` to avoid confusion. 

In this instance I would suggest adding an `odin_timestamp` column, equivalent to your existing `ROW_TIMESTAMP`. And instead of `ROW_DELETED` something like `odin_most_recent` as flag field containing either `NULL` or `Y` (or any equivalent flag distinction). 

With this structure, anything flagged as `odin_most_recent` was provided in the last API call and all other records are considered history. This would make pulling the latest data (for update operations) consistently performant as only one filtering action would be required.

src/odin/spare/spare_job.py
```diff
@@ -0,0 +1,383 @@
...
+    """
+    iso8601 = path.split("/")[-1].split("_")[0]
+    if not iso8601.endswith("Z"):
+        raise Exception(
+            f'Expected iso8601 timestamp with a Z. Got "{iso8601}". From path "{path}".'
+        )
+    return int(isoparse(iso8601).timestamp() * 1000)
+
+
+def merge_new_data(
```
rymarczy, line 238: I think this operation expects a parquet file that will always be able to fit into the system memory of the ECS.

How large do you expect this parquet file to get? Depending on the number of expected rows being added per day, the file could become too large to materialize in memory.

src/odin/spare/spare_job.py
```diff
@@ -0,0 +1,383 @@
...
+        return
+
+    df = load_parquet(parquet_path, columns)
+    changes_in_any_file = False
+    for path in paths:
+        with spare_s3.open_file(os.path.join(source, path)) as f:
+            input_df = load_json(f, columns)
+        # add metadata
+        datetime_ms = datetime_from_delta_path_ms(path)
+        input_df = add_to_left(
```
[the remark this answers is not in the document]
skyqrose: I tried that first. That adds the columns to the right. Column order isn't too important here, but these columns felt better to add to the left. For three reasons:
- Different api endpoints will have different columns but the same metadata columns, and this way they all have the shared column in the same place (col 1 and col 2).
- If any new columns are added to the API later, they'd be added to the right of this table, and I wouldn't want the metadata to end up in the middle.
- I always find it annoying when looking at wide tables and the column I care about requires a horizontal scroll to find instead of being visible at the left. I'd rather have these columns immediately visible than some random data column like `ownerUserId`.

skyqrose (commented):
Something I think I didn't explain very well in the module doc or the PR description: There's about 700 vehicles. And this first api endpoint just contains static data about them, not realtime locations. So it's not updating frequently. A new row would only be added if someone edits the vehicle's properties or adds a new vehicle, so I expect a couple thousand rows in this file ever.

And I plan to reuse this for other rarely updated static information that Spare exposes in the API, like policies, zones. Maybe Riders+Drivers (which each would be more numerous but still each one is updated rarely). Live updating stuff like locations would be handled in a different way that I haven't figured out yet (probably more like LAMP does it).

As I add more endpoints over time, I'll keep in mind the resource usage. Do you have a rule of thumb about when data is too big to fit in memory and needs to be processed with streaming?

Logging: I'll take another pass through to add some logging.

Errors: Ah, that's good to learn, thanks. I'll probably work a try/catch into the logging but otherwise won't try to recover.

src/odin/spare/spare_job.py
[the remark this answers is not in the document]
rymarczy: adding to the left would be:

```
input_df = input_df.select(
    pl.lit(datetime_ms, pl.Datetime("ms")).alias("ROW_TIMESTMAP"),
    pl.lit(False, pl.Boolean()).alias("ROW_DELETED"),
    pl.all(),
)
```

src/odin/spare/spare_job.py
```diff
@@ -0,0 +1,383 @@
...
+    Path may start with s3:// or be a local filesystem file.
+    If the file doesn't exist, return an empty DataFrame with the columns from the schema.
+    """
+    columns = METADATA_COLUMNS | columns
+    # TODO assert that loaded file has same columns in same order
+    # TODO save the file timestamp to catch conflicts
+    if parquet_path.startswith("s3://"):
+        if s3.object_exists(parquet_path):
+            s3_stream = s3.stream_object(parquet_path)
+            return pl.read_parquet(s3_stream)
```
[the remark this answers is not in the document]
rymarczy: I believe if your aws credentials are stored in a `.aws` folder in your home dir, then this should just work. I guess it could depend on your individual development setup, but this has always just worked for me...

src/odin/spare/spare_job.py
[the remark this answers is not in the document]
rymarczy: However, this is an anti-pattern as inserting at the front of a data-frame is much less efficient than appending to the end.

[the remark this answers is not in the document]
skyqrose: Ah, thanks, I like that select call better. 8eea7f9

It should be able to reuse the data buffers, so no matter which way it's concatenated, only the metadata would have to be rewritten. (It can rewrite the buffers if you do `concat(rechunk=True)` ([docs](https://docs.pola.rs/api/python/stable/reference/api/polars.concat.html)), but that's not the default.)

And this comes right after parsing JSON into a new dataframe, so any performance impact would be unobservably small compared to that.

src/odin/spare/spare_job.py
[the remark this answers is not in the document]
skyqrose: That won't work during unit tests, where I don't want to hit the real s3. For our own s3 calls in unit tests, we make a mocked s3 client, but I can't figure out how to get polars to use that client instead of making requests with my real credentials.

tests/spare/spare_job_test.py
skyqrose: Added docstrings in [c575e2a](https://github.com/mbta/odin/pull/20/commits/c575e2a3924cddfba6e2906c525447ddd42ccfec)

src/odin/spare/spare_job.py
rymarczy: What I meant by "essentially-unreadable" was related to a hypothetical situation. If this job was schedule to run every minute, and would be receiving new data every minute, it would hypothetically be re-writing the output parquet file every minute. This would create a race condition with anything reading the parquet file. Readers of the parquet file would have very frequent read errors, making the file "essentially-unreadable". 

Based on the additional information that has been provided, regarding the dimensions of this dataset, something around a 6 hour (`next_run_secs=21600`) update frequency seems more appropriate. 

The current design goals of Odin are not intended to support sub 1 hour update frequencies.

skyqrose: Okay. Eventually I do want to find a way to go more frequently, since that's something OTA has specifically asked for, but every 6 hours should be fine for now and for the medium term. I've changed it to once per 6 hours [d4b79e2](https://github.com/mbta/odin/pull/20/commits/d4b79e28c0eddf42ff75202ab38fd9c05777c506) , and I've made an Asana task to followup later (potentially after the launch) about pushing the frequency higher https://app.asana.com/1/15492006741476/project/1209584755824899/task/1210176270361205 .

src/odin/spare/spare_job.py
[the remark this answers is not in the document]
rymarczy: Could you provide some more detail on what specifically isn't working? For unit testing you should be able to mock local paths instead of s3:// paths. Extensive mocking of s3 functions/clients can lead to it's own issues, so I avoid it whenever possible.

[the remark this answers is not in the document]
skyqrose: I debugged it a bit more, and ended up getting unit tests to work by mocking the polars read/write_parquet calls, so that polars never gets a chance to hit s3. b7c1d3a (And you're right, when running the app when not testing, it just works.)

And yeah, it's kind of a pain to mock s3 calls in unit tests, but I did want at least one test that checks that the s3 code paths work, and then most of the tests use local files.

src/odin/ingestion/spare/spare_job.py
```diff
@@ -0,0 +1,400 @@
...
+    """
+    iso8601 = path.split("/")[-1].split("_")[0]
+    if not iso8601.endswith("Z"):
+        raise Exception(
+            f'Expected iso8601 timestamp with a Z. Got "{iso8601}". From path "{path}".'
+        )
+    return int(isoparse(iso8601).timestamp() * 1000)
+
+
+def merge_new_data(
```
rymarczy, line 233: Suggestion: are the `created_at` and `updated_at` fields available in this dataset reliable? If so, those fields, in combination with an `id` field, could simplify this process.

src/odin/ingestion/spare/spare_job.py
```diff
@@ -0,0 +1,400 @@
...
+        # update is newer than existing data
+        # (if files are processed out of order, an old file shouldn't
+        # uncreate a row that was just made in a previously-prcoessed new file.)
+        (pl.col("ROW_TIMESTAMP") < datetime_ms)
+    ).get_column("id")
+    new_and_changed_ids = joined.filter(
+        # if any column is updated.
+        # (this also includes wholly new rows, which have an old id of None,
+        # and undeleted rows, which have old ROW_DELETED of true
+        pl.any_horizontal(pl.col(col).ne_missing(pl.col(col + "_new")) for col in columns)
```
rymarczy, line 265: Question: Does this operation work with `list` or `struct` type columns. The `accessibilityFeatures` column of this dataset may not work with these types of operations.

src/odin/ingestion/spare/spare_job.py
```diff
@@ -0,0 +1,400 @@
...
+    should_write_parquet = rows_after > rows_before
+    ProcessLog(
+        "SpareJob write_parquet",
+        api_name=api_endpoint["name"],
+        write_parquet=should_write_parquet,
+        rows_before=rows_before,
+        rows_after=rows_after,
+        parquet_path=parquet_path,
+    )
+    if should_write_parquet:
```
rymarczy, line 355: It looks like if you end up processing files where no records change, and only `ROW_TIMESTMAP` is updated, they won't be written back out to the parquet file.

src/odin/ingestion/spare/spare_job.py
```diff
@@ -0,0 +1,400 @@
...
+from odin.job import job_proc_schedule
+import odin.utils.aws.s3 as s3
+
+
+class Config(TypedDict):
+    """Job config from config.toml, with paths"""
+
+    source: str
+    target: str
+    archive: NotRequired[str]
```
rymarczy, line 72: Concern: This configuration design obfuscates S3 bucket names and partitions in variables that need to be pulled from another repo. In the future, this will likely make modifying or debugging this process more difficult as no other Jobs use this design pattern.

rymarczy (commented):
Few more comments/suggestions

src/odin/ingestion/spare/spare_job.py
skyqrose: Yeah, that's intentional. Most of the time, most data in this dataset isn't changing. Skipping the write means the file won't grow larger with rows that don't contain new information. The tradeoff is that there's no timestamp in the file indicating how long it's been since it polled for updates, but I think that's fine.

src/odin/ingestion/spare/spare_job.py
skyqrose: The bucket name needs to be pulled from another repo, because it's different in the dev and prod deployments, so it has to come from an environment variable.

I would rather have the prefix stored alongside the bucket name, even if it's in a different repo. This makes it easier to see which paths are used in which buckets, as opposed to assuming that every bucket is organized the same way with the same paths. Having a standalone prefix name without a bucket is not very meaningful. An example of a modification that is easier to do with the bucket and prefix grouped together is described here: https://github.com/mbta/devops/pull/2845#discussion_r2075667132

An alternative would be to group everything into one bucket instead of 3 (incoming/archive/springboard), then define the bucket with a single environment variable. (and use a single directory if working locally.) Then there would be no risk of mixing up which paths belong in which buckets, and the relative prefixes within that bucket could be defined in source code.

skyqrose:
The conversation about the metadata fields I'm adding made me realize that I was assuming that saving history without overwriting anything would be necessary, but I never actually validated that with OTA. This is a stakeholder requirements question, so I can't answer it myself. I have a meeting with them Thursday to discuss.

One outcome of that meeting could be that it's not necessary, I can change this script to completely replace the parquet file with the latest data, and everything gets a lot simpler. Another outcome could be that it is necessary, and then I still need to resolve all these complex issues about detecting changes.

I won't know the answer until Thursday, so I'll let this sit until then.

src/odin/ingestion/spare/spare_job.py
skyqrose: Just tested `ne_missing` for lists of structs and it works as expected:
```py
import polars as pl
coltype = pl.List(pl.Struct({"field": pl.String()}))
df = pl.DataFrame([
  {"a": [],  "b": []},
  {"a": [{"field": "value"}],  "b": [{"field": "value"}]},
  {"a": [{"field": "value"}],  "b": [{"field": "value2"}]},
  {"a": [{"field": "value"}],  "b": []},
  {"a": None,  "b": []},
], {"a": coltype, "b": coltype})
df.select(pl.col("a").ne_missing(pl.col("b")))
# false, false, true, true, true
```

rymarczy (approved):
:cake:

skyqrose:
After talking to OTA, a lot of this metadata for saving history isn't necessary. So instead of addressing all of the comments related to it, I've cut a bunch of it out.

It's likely that I'll want to take some pieces of this approach in the future if I want to save history to other Spare API endpoints, so I'm closing this PR to preserve its current state, and I made the big simplification in a new PR #30. There's a better description of the changes I've made over there.

PR #23, opened 2025-05-06
FIX: Drop manual /tmp folde cleanup in all Jobs

rymarczy:
Reverting the addition of the `delete_folder_contens` (spelling) function.

This change resulted in some kind of race condition with the regular `/tmp` folder cleanup process. Shelving this idea for now, but need to re-visit if a specific job failure mode causes disk space issues.

### DIFF
--- a/src/odin/job.py
+++ b/src/odin/job.py
@@ -10,7 +10,7 @@ import sched
 from odin.utils.logger import ProcessLog
 from odin.utils.logger import MdValues
 from odin.utils.runtime import sigterm_check
-from odin.utils.runtime import delete_folder_contens
+from odin.utils.runtime import delete_folder_items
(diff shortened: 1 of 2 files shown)


huangh (approved):
maybe set ignore_errors would alleviate? 

shutil.rmtree(path, ignore_errors=False, onerror=None, *, onexc=None, dir_fd=None)

rymarczy:
> maybe set ignore_errors would alleviate?
> 
> `shutil.rmtree(path, ignore_errors=False, onerror=None, *, onexc=None, dir_fd=None)`

I think the error is actually occurring here:
https://github.com/mbta/odin/blob/35bee31fc1cc6a3bcb7b46261d2c925a10df432f/src/odin/job.py#L28

PR #26, opened 2025-05-12
feat: AFC API `count` endpoint fixed.

rymarczy:
The S&B AFC API `count` endpoint has been fixed to always return a limited number of results.

This changed has allowed a full featured pagination implementation for pulling data from the API.

The partial pagination solution has been removed from this process and the archive process now assumes that full pagination is always possible.

### DIFF
--- a/src/odin/ingestion/afc/afc_archive.py
+++ b/src/odin/ingestion/afc/afc_archive.py
@@ -27,12 +27,6 @@ import polars as pl
 NEXT_RUN_DEFAULT = 60 * 60 * 6  # 6 hours
 
 API_ROOT = "https://dwhexperianceapi-production.ir-e1.cloudhub.io/api/v1/datawarehouse"
-# This API_ROOT bypasses a load balancer, allowing for long running requests.
-# This is required because the API currently does not have a complete pagination solution.
-# This api endpoint is also using a self-signed cert, so SSL verifcation is not posssible.
-API_ROOT = (
-    "https://mule-worker-dwhexperianceapi-production.ir-e1.cloudhub.io:8082/api/v1/datawarehouse"
-)
 
 # Table returned by `tableinfos` endpoint as of April 30, 2025
 API_TABLES = [
@@ -81,12 +75,7 @@ def make_pl_schema(schema_list: ApiSchema) -> pl.Schema:
 
 
 class ArchiveAFCAPI(OdinJob):
-    """
-    Combine AFC API files into single parquet file.
-
-    This process should not be deployed to PROD. It has no pagination capability so resource
-    usage of this process is completely un-knowable.
-    """
+    """Combine AFC API files into single parquet file."""
 
     def __init__(self, table: str) -> None:
         """Create Job instance."""
@@ -126,12 +115,15 @@ class ArchiveAFCAPI(OdinJob):
 
     def setup_job(self) -> None:
         """
-        Grab API table schema and largest SID already processed.
+        Gather data required to run ArchiveAFCAPI.
 
-        Use of SID is used as a "pagination" method of API. One SID worth of records is the minimum
-        number of records that can be pulled at one time from the API, however there is no control
-        over how many records can/will be in one SID.
+        1. Pull Table schema from `tableinfos` endpoint.
+        2. Capture largest sid already processed from parquet dataset.
+        3. Pull list of available sid's from `count` endpoint.
+            `count` endpoint should always return some amount of sid's, even when called with only
+            "table_name" parameter.
         """
+        log = ProcessLog("setup_job", table=self.table)
         # set self.schema
         tableinfos_url = f"{API_ROOT}/tableinfos"
         r = self.make_request(tableinfos_url)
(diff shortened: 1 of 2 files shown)


src/odin/ingestion/afc/afc_archive.py
```diff
@@ -203,37 +185,33 @@ def load_sids(self) -> None:
...
-        to be pulled from `count` endpoint, however as of May 1, 2025 `count` endpoint is not
-        functional with the use of `sidFrom` parameter.
+        SID pagination will be done with data returned from `count` endpoint.
 
-        If pagination is possible, API data will be saved in batches, targeting file size of
-        approximately 10mb.
+        API data will be saved in batches.
         """
-        if self.job_ids is None:
-            self.download_csv(self.last_sid)
```
rymarczy, line 214: Never downloading files with unknown SID range anymore.

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -143,56 +135,46 @@ def setup_job(self) -> None:
         else:
             raise IndexError(f"{self.table} not found in 'tableinfos' API endpoint.")
 
-        # set self.last_sid
-        self.last_sid = None
+        # set self.pq_sid from parquet dataset
+        self.pq_sid: int | None = None
+        self.max_sid: int | None = None
```
rymarczy, line 140: `self.max_sid` only purpose is to determine if job will need immediate re-run, or not.

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -143,56 +135,46 @@ def setup_job(self) -> None:
...
-            _, self.last_sid = ds_metadata_min_max(
-                ds_from_path(f"s3://{self.export_folder}"), "sid"
-            )
+            _, self.pq_sid = ds_metadata_min_max(ds_from_path(f"s3://{self.export_folder}"), "sid")
+        log.add_metadata(pq_sid=self.pq_sid)
 
-        # set self.job_ids (if possible)
+        # pull list of sid's available to process API
         count_url = f"{API_ROOT}/count"
-        try:
```
rymarczy, line 156: no try/except needed now that this should always return valid list

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -288,35 +262,56 @@ def sync_parquet(self) -> None:
...
                 - pulling table schema from API Endpoint
-                - grab the largest SID already loaded from parquet files
-                - attempt to grab SID's available from /count endpoint
-            2a. Download all available SID's as csv.gz glob.
-                - No true pagination available for this process.
-                - Use of "sidFrom" param is inclusive of SID submitted.
-            2b. Loop through all available SID's returned from /count endpoint.
-                - Download each SID as a csv.gz file.
+                - grab the largest SID already processed, from parquet files
+                - grab SID's available from /count endpoint
```
huangh, line 291: (Question) Is there a possible error case/detection for when the /count endpoint returns available SIDs that are not the expected set of SIDs? Does this matter to us if there's a chance of re-processing already processed or skipping SIDs

huangh: Now that I'm reading the above logic/comments again it feels like this might not ever be a concern.

huangh (approved):
🙆‍♂️

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -203,37 +185,33 @@ def load_sids(self) -> None:
...
+            end_sid = job_sid
+            self.max_sid = job_sid
+            batch_count += int(api_job["dataCount"])
+            if batch_count > target_rows:
                 self.download_csv(start_sid, end_sid)
+                batch_count = 0
+                start_sid = -1
+            # stop if disk is getting too full
+            if disk_free_pct() < 60:
+                break
```
huangh, line 212: (suggestion): maybe some logging here so you know if this case hits?

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -203,37 +185,33 @@ def load_sids(self) -> None:
...
-                # stop if disk is getting too full
-                if disk_free_pct() < 60:
-                    break
-            if batch_count > 0:
+        batch_count = 0
+        # assume 12 bytes per column to ballbark 20mb per csv request
+        target_rows = 20 * 1024 * 1024 / (12 * self.schema.len())
+        start_sid = -1
+        end_sid = -1
+        for api_job in sorted(self.job_ids, key=itemgetter("jobId")):
```
huangh, line 197: do you need to be updating the sid on each iteration of the loop here? or should you just iterate over `enumerate(sorted(self...`, and then set all the start/end/max_sid outside of the loop?

src/odin/ingestion/afc/afc_archive.py
rymarczy: I don't think we really care why we move on to the `sync_parquet` part of the process. Every log statemented does have a `disk_free_mb` printed, so we will implicitly know if disk availability is decreasing. 

And then if there are more SID's available to process, the job will be triggered to run again in 5 mins.

src/odin/ingestion/afc/afc_archive.py
rymarczy: Per S&B's description of this process, SID's are guaranteed to be sequential. This process does rely on trusting that statement, and I don't know if there's necessarily a good way to confirm the process is working like that. 

The design of this process is that it will skip any SID values that are lower than the max SID of whatever we have already processed. So there shouldn't be a way to re-processing already processed data. 

If this count endpoint does not work as expected, we would have to notify S&B of the error in their process and most likely reset all of the S3 archive datasets.

src/odin/ingestion/afc/afc_archive.py
rymarczy: `self.job_ids` is expected to having the following structure:
```
List[
  {"jobId":1, "dataCount": 10},
  {"jobId":2, "dataCount": 20},
  {"jobId":3, "dataCount": 15},
]
```

In this structure, `jobId` is equivalent to `sid` (great design). 

So the sorted call on the `jobId` key is just to make certain that all the jobs are sorted by `sid` in ascending order. They are usually received like that from the API, but this is just to make certain, as it's central to the process. 

We can't set the `start_sid`/`end_sid` values outside of the loop because we're only keep tracking of those values for our call to `self.download_csv()` which can happen many times inside of the loop.  

`self.max_sid` is just meant to be save the last `sid` that was downloaded so we can check if additional `sid`/`jobId` are available at the end of the process for an immediate re-scheduling of the job.

huangh (approved):

src/odin/ingestion/afc/afc_archive.py
huangh: I can envision a more compact version of this without so many temp variables floating around (with cumsum(item_count) % target_rows...), but it's deeply in the realm of (nit).

PR #29, opened 2025-05-15
fix: AFC API File Download Stream

rymarczy:
This change updates the `.download_csv()` method of ArchiveAFCAPI class to correctly stream the file to disk.

Previous implementation was still reading entire response into memory so it could be uncompressed and written to csv file in one step.

Updated process first streams the csv.gz download to a gz file, and then uncompresses the .gz file to .csv in a seperate step.

This should resolve all memory issues associated with downloading potentially large csv.gz files.

### DIFF
--- a/src/odin/ingestion/afc/afc_archive.py
+++ b/src/odin/ingestion/afc/afc_archive.py
@@ -1,8 +1,8 @@
 import os
 import gzip
 import sched
+import shutil
 import urllib3
-from io import BytesIO
 from operator import itemgetter
 from pathlib import Path
 from typing import Literal
@@ -103,6 +103,7 @@ class ArchiveAFCAPI(OdinJob):
         :param method: (Optional) HTTP request method (such as GET, POST, PUT, etc.)
         :param fields: (Optional) Data to encode and send in request body.
         :param preload_content: (Optional) If True (default) response body preloaded into memory.
+                                If False used, call `release_conn()` when done with response
 
         :return: API Response
         """
@@ -214,7 +215,7 @@ class ArchiveAFCAPI(OdinJob):
 
         Only download csv.gz table with known sid_from and sid_to range.
 
-        :param download_jobs: list of API jobId's (sid's) to download
+        :param download_jobs: list of API jobId's (sid's) to download (sorted by jobId - ascending)
         """
         sid_from = download_jobs[0]["jobId"]
         sid_to = download_jobs[-1]["jobId"]
@@ -227,8 +228,6 @@ class ArchiveAFCAPI(OdinJob):
             num_rows=num_rows,
         )
 
-        csv_path = os.path.join(self.tmpdir, f"{sid_from}.csv")
-
         url = f"{API_ROOT}/stagetable"
         fields = {
             "table_name": self.table,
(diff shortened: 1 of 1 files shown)


src/odin/ingestion/afc/afc_archive.py
```diff
@@ -238,9 +237,22 @@ def download_csv(self, download_jobs: ApiCounts) -> None:
...
-            with open(csv_path, mode="wb") as writer:
-                writer.write(gdata.read())
+
+        csv_path = os.path.join(self.tmpdir, f"{sid_from}.csv")
+        gz_path = f"{csv_path}.gz"
+
+        # download csv.gz file (as stream)
+        with open(gz_path, mode="wb") as gz_write:
+            shutil.copyfileobj(r, gz_write)
+        r.release_conn()
```
rymarczy, line 247: Actually read the urllib3 docs: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#streaming-and-io

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -238,9 +237,22 @@ def download_csv(self, download_jobs: ApiCounts) -> None:
...
-        with gzip.open(BytesIO(r.data)) as gdata:
-            with open(csv_path, mode="wb") as writer:
-                writer.write(gdata.read())
+
+        csv_path = os.path.join(self.tmpdir, f"{sid_from}.csv")
+        gz_path = f"{csv_path}.gz"
+
+        # download csv.gz file (as stream)
+        with open(gz_path, mode="wb") as gz_write:
+            shutil.copyfileobj(r, gz_write)
```
rymarczy, line 246: By default, `copyfileobj` will loop over the response data in chunks to avoid uncontrolled memory consumption. https://docs.python.org/3/library/shutil.html#shutil.copyfileobj

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -238,9 +237,22 @@ def download_csv(self, download_jobs: ApiCounts) -> None:
...
+        r.release_conn()
+
+        # convert to csv.gz to csv, so polars scan_csv can be used
+        # csv file should also produce more consistent disk and memory usage profiles
+        with gzip.open(gz_path, mode="rb") as gz_read:
+            with open(csv_path, mode="wb") as csv_write:
```
huangh, line 252: 1) If some how you run out of disk space, I think these would exception and you'd never remove the gz_path. But it sounds like from the other PRs i've read the only way you get these is by range, so it might be worth it to be more defensive here to make sure your processed files stay in order through exceptions via putting the whole operation in a `try open/except err/finally remove gz_path`

2) doesn't save any characters, but nested `with open` feels...sus. probably better that they're both managed by the same context manager..

```
with zip.open(gz_path, mode="rb") as gz_read, open(csv_path, mode="wb") as csv_write:
    shutil.copyfileobj(gz_read, csv_write)

```

rymarczy: If running out of disk space throws a regular exception, the `OdinJob` cleanup will catch that, reset the tmpdir, and clear out any existing files from the process. 

This is pretty unlikely, as we're checking on every download for more than 40% disk usage. It would really only happen if a single file ends up being about 50% of the entire disk space.

huangh (approved):
🖌️

PR #30, opened 2025-05-15
feat(spare): process Spare vehicles data into Parquet (simplified version)

skyqrose:
Simplified version of #20 

Asana Task: [Process Spare Dev data into Parquet with Odin](https://app.asana.com/1/15492006741476/project/1209584755824899/task/1209698688134966?focus=true)

Most of the back and forth on #20 was related to the metadata fields and the merging logic. When trying to address some of the feedback, I realized a lot of that complexity might not be needed, at least for this data. I checked with OTA, and saving a historical record of what the vehicle data was at any point in the past isn't important. So instead of fixing it, I've deleted a bunch of it instead.

This now takes a much simpler approach of converting the latest input directly into parquet. Anything but the latest input file is ignored. Any old parquet file is completely ignored and overwritten. The parquet file will then contain exactly the same data that's currently in Spare, and no more.

If it helps for reviewing, here's the diff from #20: https://github.com/mbta/odin/compare/d1fe68d..sqr-spare2 The only code to re-review is in `spare_job.py` and `test_spare_job.py`.

It's almost certain that some other data from Spare will require saving history when I add support for more api endpoints in the near future, but it may not end up looking exactly like this. So I think it'll still be good to remove it for now and only add history back once I have a better sense of what it needs to be.


rymarczy (approved):
I'm not super familiar with the capabilities of delta, however if the goal is to just write the latest of these API results to a location on S3, delta is already doing that. 

If writing only the `data` element of the API results to an s3 location is a requirement, and it's not possible to accomplish that with delta, then this all looks good to me.

PR #31, opened 2025-05-19
feat: Automatic Odin Data Dictionary Export

rymarczy:
This change adds a job to automatically create data dictionary exports for all parquet datasets in the SPRINGBOARD/ODIN_DATA partition.

This is just an initial draft of the data dictionary format and will likely change as needs arise.

### DIFF
--- a/src/odin/generate/data_dictionary/dictionary.py
+++ b/src/odin/generate/data_dictionary/dictionary.py
@@ -1,0 +1,116 @@
+import os
+import json
+import sched
+from dataclasses import dataclass
+from dataclasses import asdict
+from typing import Generator
+from typing import Any
+
+import pyarrow as pa
+import pyarrow.parquet as pq
+
+from odin.job import OdinJob
+from odin.job import job_proc_schedule
+from odin.utils.locations import DATA_SPRINGBOARD
+from odin.utils.locations import ODIN_DATA
+from odin.utils.locations import ODIN_DICTIONARY
+from odin.utils.aws.s3 import list_objects
+from odin.utils.aws.s3 import list_partitions
+from odin.utils.aws.s3 import upload_file
+
+NEXT_RUN_DEFAULT = 60 * 60 * 12  # 12 hours
+ODIN_ROOT = f"s3://{os.path.join(DATA_SPRINGBOARD, ODIN_DATA)}"
+
+
+@dataclass
+class DatasetField:
+    """Values exepcted for DatasetField"""
+
+    column_name: str
+    column_type: str
+    is_partition: bool
+
+    def __init__(self, column: pa.Field) -> None:
+        """Create DatasetField from pyarrow field."""
+        self.column_type = str(column.type)
+        if isinstance(column.type, pa.DictionaryType):
+            self.column_type = str(column.type.value_type)
+        self.column_name = column.name
+        # in testing DictionaryType is always partition column, futher testing may change this...
+        self.is_partition = isinstance(column.type, pa.DictionaryType)
+
+
+@dataclass
+class DatasetDictionary:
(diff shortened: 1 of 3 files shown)


src/odin/generate/data_dictionary/dictionary.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __init__(self, ds_path: str) -> None:
+        """Create DatasetDictionary from datset path."""
+        data_objects = list_objects(ds_path, in_filter=".parquet")
+        self.dataset_group = os.path.dirname(ds_path).replace(ODIN_ROOT, "").strip("/")
+        self.dataset_name = os.path.basename(ds_path)
+        self.dataset_path = ds_path
+        self.size_mb = int(sum([obj.size_bytes for obj in data_objects]) / (1024 * 1024))
+        self.child_files = [obj.path for obj in data_objects]
+        self.schema = [DatasetField(c) for c in pq.ParquetDataset(ds_path).schema]
```
huangh, line 62: looks like this is intending to autogenerate this dictionary regularly, so i'm assuming it's not adding other metadata/commentary? 

do you need it in this exact json schema format for some future purpose? because the built ins can return the data you're grabbing, with less code. 


`table.schema.names`
`table.schema.types`

or if you convert a batch to polars, `df.collect_schema()`

src/odin/generate/data_dictionary/dictionary.py
```diff
@@ -0,0 +1,116 @@
...
+from odin.job import OdinJob
+from odin.job import job_proc_schedule
+from odin.utils.locations import DATA_SPRINGBOARD
+from odin.utils.locations import ODIN_DATA
+from odin.utils.locations import ODIN_DICTIONARY
+from odin.utils.aws.s3 import list_objects
+from odin.utils.aws.s3 import list_partitions
+from odin.utils.aws.s3 import upload_file
+
+NEXT_RUN_DEFAULT = 60 * 60 * 12  # 12 hours
```
huangh, line 21: the only information that seems to get updated regularly is the `size_mb` and child files. are you keeping the schema in there because you intend for it to change regularly and you just want this to auto-document?

src/odin/generate/data_dictionary/dictionary.py
```diff
@@ -0,0 +1,116 @@
...
+    is_partition: bool
+
+    def __init__(self, column: pa.Field) -> None:
+        """Create DatasetField from pyarrow field."""
+        self.column_type = str(column.type)
+        if isinstance(column.type, pa.DictionaryType):
+            self.column_type = str(column.type.value_type)
+        self.column_name = column.name
+        # in testing DictionaryType is always partition column, futher testing may change this...
+        self.is_partition = isinstance(column.type, pa.DictionaryType)
```
huangh, line 40: this is the most curious part of the design for me...is there use case for mingling the partition names with the column names? 

i would almost rather this all get exported to a dataframe instead of json, so you can do easy filtering. but as it currently stands you'd have to convert it to a usable form each time to extract anything. a wiser reincarnation of the metadata table?

Is this for human consumption or machine consumption? (like the lamp data dictionaries..)

huangh (commented):
Have a few suggestions and questions on intent and design of the dictionary

src/odin/generate/data_dictionary/dictionary.py
rymarczy: We want it to be auto-documented as there could be hundreds or thousands of schemas that we won't want to track manually. 

The schema can also technically change, in the situation that a column is added to a later parquet file of the dataset or a dataset is "refreshed" because of a schema change operation.

huangh (approved):

PR #34, opened 2025-05-22
feat: process seperate S&B AFC API table types

rymarczy:
S&B API is now providing a `type` value from the `tableinfos` endpoint.

The `type` can either be "static" or "transactional".

"static" indicates the table should only ever have 1 sid/jobId value at a time and performs a full truncate and reload for any new data.

"transactional" is an incremental (append-only) data loading strategy.

The `ArchiveAFCAPI` Job previously assumed all tables operated as "transactional" this change updates the process to also handle "static" table types.

### DIFF
--- a/src/odin/ingestion/afc/afc_archive.py
+++ b/src/odin/ingestion/afc/afc_archive.py
@@ -20,9 +20,11 @@ from odin.utils.locations import DATA_SPRINGBOARD
 from odin.utils.aws.s3 import list_objects
 from odin.utils.aws.s3 import download_object
 from odin.utils.aws.s3 import upload_file
+from odin.utils.aws.s3 import delete_objects
 from odin.utils.parquet import ds_metadata_min_max
 from odin.utils.parquet import ds_from_path
 from odin.utils.parquet import pq_dataset_writer
+from odin.utils.parquet import ds_unique_values
 
 import polars as pl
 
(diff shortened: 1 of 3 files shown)


src/odin/ingestion/afc/afc_archive.py
```diff
@@ -325,11 +327,21 @@ def sync_parquet(self) -> None:
             return
 
         found_objs = list_objects(f"s3://{self.export_folder}", in_filter=".parquet")
-        if found_objs:
-            sync_file = found_objs[-1].path.replace("s3://", "")
-            s3_pq_file = os.path.join(self.tmpdir, sync_file.replace("/table_", "/temp_"))
-            download_object(found_objs[-1].path, s3_pq_file)
-            sync_paths.insert(0, s3_pq_file)
+        del_objs = []
+        if self.table_type == "transactional":
```
grejdi-mbta, line 331: Can you add a test for this? I only see a "static" one.

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -340,6 +352,7 @@ def sync_parquet(self) -> None:
 
         # Check for sigterm before upload (can't be un-done)
         sigterm_check()
+        delete_objects(del_objs)
```
grejdi-mbta, line 355: Also, I feel like the `if len(objects) == 0:` should be here. It's a little bit clearer that we are only deleting if there are objects.

grejdi-mbta (changes requested):

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -325,11 +327,21 @@ def sync_parquet(self) -> None:
...
-            download_object(found_objs[-1].path, s3_pq_file)
-            sync_paths.insert(0, s3_pq_file)
+        del_objs = []
+        if self.table_type == "transactional":
+            # `transactional` table type is supposed to be incremental (append-only) data model
+            if found_objs:
+                sync_file = found_objs[-1].path.replace("s3://", "")
+                s3_pq_file = os.path.join(self.tmpdir, sync_file.replace("/table_", "/temp_"))
+                download_object(found_objs[-1].path, s3_pq_file)
+                sync_paths.insert(0, s3_pq_file)
```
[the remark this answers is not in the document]
rymarczy: The last is never very large, on average probably less than 10 elements. 

Putting the S3 file at the start of this list produces an export parquet file with records in the order they were received from the API. The benefit of having the parquet file records in a logical order greatly outweighs the cost of a list insert operation.

src/odin/ingestion/afc/afc_archive.py
rymarczy: There are no issues with passing a list of 0 elements into `delete_objects`. Several other parts of the application do that regularly.

Previously, passing [] would just cause a ThreadPool to spin up, do nothing, and then shut down, with some associated logging output. Adding the `len(object)==0` check to `delete_objects` mostly just saves some unnecessary logging output.

src/odin/ingestion/afc/afc_archive.py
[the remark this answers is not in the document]
huangh: Sounds good. 

(suggestion) - feels worthy of a comment here to state the benefits, as the insert(0, blah) is kind of innocuous, but the follow-on effects seem to be important.

grejdi-mbta (approved):

PR #35, opened 2025-05-23
fix: ds_from_path schema discovery

rymarczy:
This change fixes an issue with the `ds_from_path` parquet utility function. When a partition path is passed into the function, that contains a large number of nested parquet files, the full dataset schema is not properly merged.

To fix this issue, the dataset created from `ds_from_path` is now made from a list of all individual parquet files in the partition. This correctly discovers the merged dataset schema.

This change also updates the cast operation inside of ds_metadata_limit_k_sorted. A cast operation was previously required to align partition and NULL columns between row group tables and dataset schema. The correct column types are now pulled direction from the dataset and applied during column creation.


huangh (approved):
🥼

PR #38, opened 2025-05-29
feat: add cubic report material views to duckdb

rymarczy:
This change adds a few new reports to the `cubic_reports` schema of the duckdb file.

Some of the reports are "material views" that are pre-compiled and saved to S3 to avoid long query times.

### DIFF
--- a/src/odin/generate/data_dictionary/cubic_reports_sql.py
+++ b/src/odin/generate/data_dictionary/cubic_reports_sql.py
@@ -102,6 +102,46 @@ COMP_B_VIEW = """
     ;
 """
 
+COMP_C_VIEW = """
+    DROP VIEW IF EXISTS cubic_reports.wc700_comp_c;
+    CREATE VIEW cubic_reports.wc700_comp_c
+    AS
+    SELECT
+        t.settlement_day_key
+        ,t.operating_day_key
+        ,rc.rider_class_name
+        ,fp.fare_prod_name
+        ,t.service_type_id
+        ,t.fare_rule_description
+        ,t.recovery_txn_type
+        ,sum(t.minimum_fare_charge) / 100 AS recovery_calculation_amount
+    FROM
+        cubic_ods.edw_farerev_recovery_txn t
+    LEFT JOIN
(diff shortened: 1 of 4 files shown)


src/odin/generate/data_dictionary/cubic_reports_sql.py
```diff
@@ -102,6 +102,46 @@
     ;
 """
 
+COMP_C_VIEW = """
+    DROP VIEW IF EXISTS cubic_reports.wc700_comp_c;
+    CREATE VIEW cubic_reports.wc700_comp_c
+    AS
+    SELECT
+        t.settlement_day_key
```
huangh, line 110: Drawing from software best practices to have..somewhat descriptive variable names, can these aliases be spelled out a bit more to help with readability?

src/odin/generate/data_dictionary/cubic_reports_sql.py
```diff
@@ -102,6 +102,46 @@
...
+    LEFT JOIN
+        cubic_ods.edw_rider_class_dimension rc
+        ON
+            rc.rider_class_id = t.rider_class_id
+    LEFT JOIN
+        cubic_ods.edw_fare_product_dimension fp
+        ON
+            fp.fare_prod_key = t.fare_prod_key
+    WHERE
+        fp.monetary_inst_type_id = 2
```
huangh, line 129: describe type_id = 2 with a comment?

huangh (commented):
Just some requests to add some comments and aid in readability.

src/odin/generate/data_dictionary/cubic_reports_sql.py
huangh: and any other magic identifier in the query

rymarczy: We don't know exactly what any of these queries or magic identifiers mean. The queries are provided by Cubic with any context.

grejdi-mbta:
@rymarczy Will you be able to do these in the DMAP-Import database as well?

PR #41, opened 2025-06-03
feat: pull job_id's from /count endpoint as generator

rymarczy:
This change modifies the way job_id's are pulled from the S&B AFC API /count endpoint. 

Previously, each Job instance pulled 1000 job_id's per run. The new implementation will continue to pull job_ids as a generator until the hard disk is filled up or there are no more job_ids to pull.

### DIFF
--- a/src/odin/ingestion/afc/afc_archive.py
+++ b/src/odin/ingestion/afc/afc_archive.py
@@ -316,6 +316,7 @@ class ArchiveAFCAPI(OdinJob):
         API data will be downloaded in batches to limit API reponse time.
         """
         # assume 12 bytes per column to ballbark 50mb per request
+        min_disk_free_pct = 60
         target_rows = int(50 * 1024 * 1024 / (12 * self.schema.len()))
         download_jobs: APICounts = []
         download_rows = 0
@@ -332,13 +333,13 @@ class ArchiveAFCAPI(OdinJob):
                     download_jobs = []
                     download_rows = 0
                 # stop if disk is getting too full
-                if disk_free_pct() < 60:
+                if disk_free_pct() < min_disk_free_pct:
                     break
             if len(download_jobs) > 0:
                 self.download_json(download_jobs)
 
             # This is ugly, but should work for now...
-            if disk_free_pct() < 60:
+            if disk_free_pct() < min_disk_free_pct:
                 break
 
     def sync_parquet(self) -> None:
--- a/src/odin/ingestion/afc/afc_archive.py
+++ b/src/odin/ingestion/afc/afc_archive.py
@@ -30,7 +30,7 @@ from odin.utils.parquet import ds_unique_values
 import polars as pl
 
 
-class APIEmptyResponse(Exception):
+class APIEmptyResponseError(Exception):
     """API Returned 204."""
 
     pass
@@ -184,7 +184,7 @@ class ArchiveAFCAPI(OdinJob):
         """
         r = self.req_pool.request(method, url=url, fields=fields, preload_content=preload_content)
         if r.status == 204:
-            raise APIEmptyResponse("API 204 Response, NO DATA.")
+            raise APIEmptyResponseError("API 204 Response, NO DATA.")
         elif r.status != 200:
             raise urllib3.exceptions.HTTPError(
                 f"API ERROR: {url=} status={r.status} response_data={r.data.decode()}"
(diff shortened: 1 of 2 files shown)


src/odin/ingestion/afc/afc_archive.py
```diff
@@ -286,25 +297,30 @@ def load_job_ids(self) -> None:
...
         download_jobs: APICounts = []
         download_rows = 0
-        for api_job in sorted(self.job_ids, key=itemgetter("jobId")):
-            if self.pq_job_id is not None and int(api_job["jobId"]) <= int(self.pq_job_id):
-                continue
-            download_jobs.append(api_job)
-            download_rows += int(api_job["dataCount"])
-            self.max_job_id = int(api_job["jobId"])
-            if download_rows > target_rows:
+        for job_ids in self.api_job_ids():
```
rymarczy, line 304: Adds `job_id` generator as outer loop. And inner loop is un-changed.

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -286,25 +297,30 @@ def load_job_ids(self) -> None:
...
+                if disk_free_pct() < 60:
+                    break
+            if len(download_jobs) > 0:
                 self.download_json(download_jobs)
-                download_jobs = []
-                download_rows = 0
-            # stop if disk is getting too full
+
+            # This is ugly, but should work for now...
             if disk_free_pct() < 60:
```
rymarczy, line 322: Extra breakout needed if loop was broken because of disk space.

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -28,6 +29,13 @@
 
 import polars as pl
 
+
+class APIEmptyResponse(Exception):
```
rymarczy, line 33: Offer different exception for  204 response "no data"

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -214,23 +224,36 @@ def setup_job(self) -> None:
...
+        while True:
+            if job_id_from is not None:
+                req_fields["jobIdFrom"] = str(job_id_from + 1)
+            try:
+                r = self.make_request(count_url, fields=req_fields)
+            except APIEmptyResponse:
+                break
+            api_counts: APICounts = r.json()
+            api_counts = sorted(api_counts, key=itemgetter("jobId"))
+            if api_counts[-1]["jobId"] != job_id_from:
```
rymarczy, line 252: This probably will never be reached with the new logic to add 1 for the `jobIdFrom` parameter.

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -214,23 +224,36 @@ def setup_job(self) -> None:
...
+                yield api_counts
+            else:
```
rymarczy, line 254: This `else` probably will never be reached with the new logic to add 1 for the jobIdFrom parameter. We may only ever break on the `APIEmptyResponse` exception.

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -409,8 +437,9 @@ def run(self) -> int:
         )
         self.setup_job()
         self.load_job_ids()
+        next_run_duration = self.re_run_check()
```
rymarczy, line 440: Some tables seem to generate new job_id's very quickly, so do the `re_run_check` right after downloading of job_ids has completed.

huangh: also i assume this is not based on reality - but there's no direct connection between the queried # job_ids (10k at a time) and the duration of wait here. so if something was generating "new job_id's very quickly" - > 10k per 5 minutes, you would fall behind, yes?

rymarczy: 5 minutes is good for an "immediate" re-run vs the default of 6 hours. 

the 10k `limit` parameter for job_id's is per request. The new generator can issue and unlimited number of requests, so millions of job_id's could be processed during a single job run, as long as the 40% disk space usage limit is not hit.

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -212,25 +222,44 @@ def setup_job(self) -> None:
...
+            # however this is not documented so this process is designed to work with the endpoint
+            # being either inclusive or exclusive of the `jobIdFrom` parameter
+            api_counts = sorted(
+                [j for j in api_counts if j["jobId"] > job_id_from], key=itemgetter("jobId")
+            )
+            if len(api_counts) > 0:
+                yield api_counts
+            else:
+                break
+            job_id_from = api_counts[-1]["jobId"]
```
huangh, line 262: This updates self.pq_job_id by reference, yes? Can we do it explicitly instead? this is a bit confusing, i was looking for where pq_job_id got incremented

huangh: either that or update the comment to state that this is a side effect/return type of this method

rymarczy: I dropped self.pq_job_id from this class method. It now has a `jod_id_from` param that just gets updated inside of the function.

huangh: i read a thing wrong, i got it now.

huangh (approved):
🥇

PR #43, opened 2025-06-13
feat: extra AFC API verification

rymarczy:
The change adds a couple extra checks to the `verify_downloads` function. 

These checks will compare the schema from the `tableinfos` endpoint to the columns found in the downloaded json file. Any differences in schema will raise an error. This is based on s&b not aligning the `tableinfos` and `stagetable` endpoint schemas.


huangh (approved):
🥇

PR #44, opened 2025-06-20
fix: use column names in duckDB views

rymarczy:
This change updates the way duckdb file VIEWS are created. VIEWS were previously created with a `SELECT *` statement, this change selects the specific columns that are expected in the parquet files.

In the situation where columns have been added to parquet files, this change will allow the existing VIEWS to still function, instead of throwing an error because the `SELECT *` statement no longer matches the metadata stored in the duckdb file.

This change also updates the duckdb version being used and addes an environment tag to duckdb and data dictionary objects.

Asana task: https://app.asana.com/1/15492006741476/project/1208949713596462/task/1210598982895512
Asana task: https://app.asana.com/1/15492006741476/project/1208949713596457/task/1210574836215977


src/odin/generate/data_dictionary/duck_db.py
```diff
@@ -12,10 +13,11 @@
 from odin.utils.locations import CUBIC_QLIK_DATA
 from odin.utils.locations import AFC_DATA
 from odin.utils.locations import CUBIC_ODS_REPORTS
+from odin.utils.parquet import ds_from_path
 import odin.generate.data_dictionary.cubic_reports_sql as cubic_sql
 
 
-DB_FILE = "fares_data_repository.db"
+DB_FILE = f"fares_data_repository_{AWS_ENV}.db"
```
grejdi-mbta, line 20: Yo. Do you mind adding this prefix only for DEV or non-PROD? It's a little cleaner.

src/odin/generate/data_dictionary/duck_db.py
```diff
@@ -28,25 +30,29 @@ class ViewBuilder:
 
 
 DROP_VIEW = "DROP VIEW IF EXISTS $schema.$table;"
-READ_PQ = "read_parquet('$s3_path/**/*.parquet')"
+READ_PQ = "read_parquet('$s3_path/**/*.parquet', union_by_name = true)"
```
grejdi-mbta, line 33: 👍  Had to do this with Masabi data I was pulling from over the last year.

grejdi-mbta (approved):

PR #45, opened 2025-06-26
feat: add WA160 view to duckdb file

rymarczy:
This change adds a view for the WA160 report to our duckdb file. This is being done at the request of the MBTA finance team.


huangh (approved):
🔢

huangh:
the duplication of query across repos bases seems like a ripe opportunity for something to get forgotten/out of sync. perhaps at least put a comment in there to make sure we keep them aligned? or do a string compare test and pull both repos at once to make sure the queries are the same?

rymarczy:
> the duplication of query across repos bases seems like a ripe opportunity for something to get forgotten/out of sync. perhaps at least put a comment in there to make sure we keep them aligned? or do a string compare test and pull both repos at once to make sure the queries are the same?  

That might be possible, but the duckdb queries in Odin are not always the exact same as the postgres queries in dmap-import. So a direct string comparison probably wouldn't work.

PR #48, opened 2025-07-01
fix: cubic REASON_DIMENSION migration

rymarczy:
This changes adds a migration to resolve a type discrepency in the EDW.REASON_DIMENSION tables.

The recent cubic back-office upgrade introcued a type change for the REASON_DIMENSION table changing the `purse_type` column from an integer to a string. To handle this type change, a new migration has been created to convert all parquet row groups to matching schemas for this table.

This change will resolve recent errors being thrown by odin because of miss-matched data types between parquet files in the same dataset.

### DIFF
--- a/src/odin/migrate/migrations/odin-dev/0005.py
+++ b/src/odin/migrate/migrations/odin-dev/0005.py
@@ -1,0 +1,68 @@
+import os
+import tempfile
+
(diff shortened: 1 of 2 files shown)


huangh (approved):
👍

PR #51, opened 2025-07-10
fix: ODSFact TRANSACTION_HISTORY OOM

rymarczy:
For the last few days, Odin has been throwing OOM errors when attempting to process the EDW.TRANSACTION_HISTORY table for the ODSFACT job.

This error was traced back to the `ds_batched_join` function, which is designed to join a polars dataframe to a very large parquet dataset in a memory constrained environment. Unfortunately, the implementation was not as memory constrained as originally thought.

In this instance, the JOIN operations was matching to a very small number of records from almost every partition across the whole dataset. In converting a pyarrow RecordBatch to a polars Dataframe, for each JOIN match, the memory allocation from each matched RecordBatch was being leaked as the dataset was iterated through.

This conversion to Polars is only required for JOIN operations on key columns that may contain NULL values. If no NULL values need to be JOIN'ed on, the JOIN operation can occur entirely in pyarrow (using Tables) and the memory leak does not occur. The NON NULL operation is the primary JOIN operation that occurs, so the Polars conversion remains for the small subset of data needing this functionality.

We'll test this on DEV to make sure all the datasets are able to execute JOIN operations correctly before deploying to PROD.

### DIFF
--- a/src/odin/utils/parquet.py
+++ b/src/odin/utils/parquet.py
@@ -114,8 +114,32 @@ def pq_rows_per_mb(source: Union[str, Sequence[str]], num_rows: Optional[int] = 
 class RowGroupStats(TypedDict):
     """Incomplete representation of Stats fields, but only ones currently used."""
 
+    has_min_max: bool
     max: Any
     min: Any
+    null_count: int
+    num_values: int
+
+
+def file_column_stats(pq_meta: pq.FileMetaData, column: str) -> list[RowGroupStats]:
+    """
+    Retrieve 'column' statistics from metadata of parquet file.
+
+    :param pq_meta: Metadata of parquet file.
+    :param column: Name of column to retrieve stats for.
+
+    :return: List of RowGroupStats, each List index correspends File RowGroup of same index.
+    """
+    col_index = pq_meta.schema.to_arrow_schema().get_field_index(column)
+    assert col_index >= 0
+    file_stats = []
+    for rg_index in range(pq_meta.num_row_groups):
+        rg_stats: RowGroupStats = (
+            pq_meta.row_group(rg_index).column(col_index).to_dict()["statistics"]  # type: ignore[assignment]
+        )
+        file_stats.append(rg_stats)
+
+    return file_stats
 
 
 def row_group_column_stats(rg_metadata: pq.RowGroupMetaData, column: str) -> RowGroupStats:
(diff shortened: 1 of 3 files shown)


src/odin/utils/parquet.py
```diff
@@ -114,8 +114,32 @@ def pq_rows_per_mb(source: Union[str, Sequence[str]], num_rows: Optional[int] =
...
     """Incomplete representation of Stats fields, but only ones currently used."""
 
+    has_min_max: bool
     max: Any
     min: Any
+    null_count: int
+    num_values: int
+
+
+def file_column_stats(pq_meta: pq.FileMetaData, column: str) -> list[RowGroupStats]:
```
rymarczy, line 124: More efficient way to pull column statistics for an entire parquet file.

src/odin/utils/parquet.py
```diff
@@ -521,31 +545,55 @@ def ds_batched_join(
...
-            raise TypeError("Always dataframe.")
-        _df = _df.cast(mod_cast).join(match_cols, on=keys, how="inner", nulls_equal=True)  # type: ignore[arg-type]
-        if _df.shape[0] > 0:
-            join_frames.append(_df)
-
-    if join_frames:
-        return_frame = pl.concat(join_frames, how="diagonal")
+
+    return_frame = pl.from_arrow(pa.Table.from_pylist([], schema=ds.schema))  # type: ignore[assignment]
+    if match_cols.null_count().sum_horizontal().item() == 0:
```
rymarczy, line 560: If no NULL values in `match_cols` Dataframe, then use the pyarrow JOIN method, otherwise the Polars method that CAN leak memory on very large datasets.

src/odin/utils/parquet.py
```diff
@@ -521,31 +545,55 @@ def ds_batched_join(
...
     else:
-        return_frame = pl.from_arrow(pa.Table.from_pylist([], schema=ds.schema))  # type: ignore[assignment]
+        # pyarrow can not perform join operations on columns containing NULL values
+        # if keys columns contain NULL values, use polars for JOIN operation
+        # converting the dataset RecordBatch to polars results in a memory leak because the memory
+        # associated with every polars DataFrame saved to join_frames is retained until the entire
+        # dataset has been iterated through
+        # this is a problem if the join operations ends up taking a very small number of records
+        # from a loarge number of paritions in a very large dataset
+        join_frames = []
```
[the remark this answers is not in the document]
rymarczy: I'm not sure if resuming this batch processing is possible. Each join operation has to occur as an all or nothing event. A memory-break-design would need to have a more complicated re-run mechanism for each time the loop is broken because of memory. 

I think only 1 or 2 tables in this dataset have keys that contain NULL values and they're very small datasets, compared to the ones doing these large join operations. 

There's definitely room for improvement here, so I've created a backlog task to re-visit this functionality in the future. https://app.asana.com/1/15492006741476/project/1208949713596462/task/1210772547162934

PR #52, opened 2025-07-15
feat: S&B API Restricted Process

rymarczy:
This change adds a new ingestion process for the S&B AFC API.

This new process removes PPI from the public version of the `v_media` table and publishes a separate "lookup" version of the table to a restricted S3 path.

This change also includes a migration to delete the existing `v_media` exports and re-pull the data with PPI removed.

Asana Task: https://app.asana.com/1/15492006741476/project/1208949713596462/task/1210707790279822

### DIFF
--- a/src/odin/ingestion/afc/afc_restricted.py
+++ b/src/odin/ingestion/afc/afc_restricted.py
@@ -1,0 +1,100 @@
+import os
+import sched
+from pathlib import Path
+
+
+from odin.job import job_proc_schedule
+from odin.utils.runtime import sigterm_check
+from odin.utils.logger import ProcessLog
+from odin.utils.locations import DATA_SPRINGBOARD
+from odin.utils.locations import AFC_RESTRICTED
+from odin.utils.aws.s3 import list_objects
+from odin.utils.aws.s3 import download_object
+from odin.utils.aws.s3 import upload_file
+from odin.utils.parquet import ds_from_path
+from odin.utils.parquet import pq_dataset_writer
+
+import polars as pl
+from odin.ingestion.afc.afc_archive import ArchiveAFCAPI
+
+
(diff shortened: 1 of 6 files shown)


src/odin/ingestion/afc/afc_restricted.py
```diff
@@ -0,0 +1,100 @@
...
+                log.add_metadata(pq_path=pq_path, pq_row_count=pq_row_count)
+
+        if len(sync_paths) == 0:
+            return
+
+        export_jobs = (
+            (self.export_folder, sync_paths),
+            (os.path.join(DATA_SPRINGBOARD, AFC_RESTRICTED, self.table), rs_sync_paths),
+        )
+        new_paths: list[str] = []
```
rymarczy, line 69: Run same "merge" process for both datasets, only difference is "export_path" and source "sync_paths)

src/odin/ingestion/afc/afc_restricted.py
```diff
@@ -0,0 +1,100 @@
...
+            if json_file.suffix != ".json":
+                continue
+            lf = pl.scan_ndjson(
+                json_file.absolute(),
+                schema=self.schema,
+            ).with_columns(pl.col(self.ts_cols).str.head(19).str.to_datetime("%Y-%m-%d %H:%M:%S"))
+
+            # create public parquet file
+            pq_path = str(json_file.absolute()).replace(".json", ".parquet")
+            lf.drop("engraveid").sink_parquet(
```
huangh, line 39: I'm about to do something similar for operator_id for BPM. Do you think its worth it to consider programmatically enforcing this at a higher level? Something like a "RESTRICTED COLUMNS" list that we populate and maintain, and then any queries/joins that grab data from the source are verified to not go into certain buckets? 

i'm more posing the question...not sure how this could be implemented with assurances

PR #59, opened 2025-08-08
chore: README and comments updates

rymarczy:
This change contains some minor README updates and context documentation for future developers.


ealexa05 (approved):
Approving; changes look good, and I understand the docs well enough to take things from here

PR #65, opened 2025-10-03
Adding override for boolean columns and migration files to regenerate…

ealexa05:
… table


ealexa05:
The basic issue here is that, while the boolean data looks fine in the raw ndjsons from the API (e.g. "valid":true and "onlineprocessingstatus":true), when polars is converting it to parquet and is given a schema in pl.scan_ndjson it is _not_ casting to match type, but instead strictly checking, so boolean values become `null` in the parquet. The onlineprocessingstatus column also has some missing values, so if no schema is supplied it falls back to float64 which allows for nulls, so an explicit schema must be provided.

There are probably other nulls in the db that stem from this String type fallback -> strict check, but this PR will address an acute issue and let us make sure there are no unintended downstream consequences before more broadly addressing schema type issues

grejdi-mbta (approved):

PR #95, opened 2026-02-09
Add retry to afc filesystem op

MaxAlex:
Asana ticket: https://app.asana.com/1/15492006741476/task/1213090243017307

This error can be caused by trying to read from incompletely written files. There's two obvious ways this could happen; either reading the file from the AFC endpoint was cut short, or the downloaded JSON gzip file wasn't completely available on disk before it was re-opened to be unzipped (much less likely, but could be due to AWS drive latency, plausibly.) In either case the error is less likely to persist if we simply re-try the download process a few times.

---
  - https://app.asana.com/0/0/1213090243017307

### DIFF
--- a/src/odin/ingestion/afc/afc_archive.py
+++ b/src/odin/ingestion/afc/afc_archive.py
@@ -4,12 +4,11 @@ import json
 import sched
 import shutil
 import urllib3
+import time
 from collections import ChainMap
 from operator import itemgetter
 from pathlib import Path
-from typing import Literal
-from typing import TypedDict
-from typing import Generator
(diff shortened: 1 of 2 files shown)


grejdi-mbta (changes requested):
Can you please add a test for this change.

ealexa05 (approved):
Changes and test look good to me! For testing, make sure that afc_archive is transferring data as expected on dev (and keep an eye out on dev to see if this issue recurs, although I think it is infrequent enough that it's not worth holding up the prod deploy as long as basic afc transfer is working)

grejdi-mbta (approved):

PR #96, opened 2026-02-23
Adding logging for python and module versions and load_cdc_records

ealexa05:
This PR adds logging in two places:

1. at start-up, Odin will log current versions of python and installed modules after checking env vars
2. in `load_cdc_records()`,  this adds extra checks to identify when updates to fact tables do not match what is expected (i.e. initial rows + inserts - deletes), and adds additional tracking along the way

Specifically for `load_cdc_records()`, this adds:

`load_cdc_initial_state`: Logs initial state of fact table. Uses `ds_column_min_max()` to get actual max value for "header__change_seq", in case it differs from what is reported in the metadata (unlikely, but possible if data is corrupted)
`load_cdc_fetch`: Logs applicable cdc results
`load_cdc_processing`: Logs cdc processing summary after loop 
`load_cdc_upload_complete`: Logs state of fact table after upload, and records `true` for `row_count_mismatch` if resulting number of rows differs from expected

These new logs should not add affect overall log volume much, nor add much processing time. After deploying, I should monitor the performance to ensure `ds_column_min_max()` is not consuming too many resources

### DIFF
--- a/src/odin/generate/cubic/ods_fact.py
+++ b/src/odin/generate/cubic/ods_fact.py
@@ -21,6 +21,8 @@ from odin.utils.locations import DATA_ARCHIVE
 from odin.utils.locations import CUBIC_QLIK_PROCESSED
 from odin.utils.parquet import fast_last_mod_ds_max
 from odin.utils.parquet import ds_metadata_min_max
+from odin.utils.parquet import ds_column_min_max
+from odin.utils.parquet import ds_files
 from odin.utils.parquet import ds_from_path
 from odin.utils.parquet import ds_unique_values
 from odin.utils.parquet import pq_dataset_writer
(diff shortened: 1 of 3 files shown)


src/odin/generate/cubic/ods_fact.py
```diff
@@ -349,6 +393,24 @@ def load_cdc_records(self) -> int:
             insert_df, update_df, delete_df = cdc_to_fact(
                 cdc_df, insert_df, update_df, delete_df, keys
             )
+
+        # Log CDC processing summary after loop
+        cdc_final_seq = cdc_df.get_column("header__change_seq").max() if cdc_df.height > 0 else None
+        process_log = ProcessLog(
+            "load_cdc_processing",
+            table=self.table,
+            loop_iterations=loop_iterations,
```
MaxAlex: Code style nit; for-loop variables aren't restricted to the loop scope, and looping over `range(10)` will always run at least once, so you could just use `loop_idx` here rather than defining `loop_iterations`.

(Not relevant to this particular change; instead of `for _ in range(10)` this ought to have been something like `while not (cdc_df.height == 0 or num_load_records > max_load_records)` , in either case.)

MaxAlex (approved):

PR #106, opened 2026-03-04
URL env var fix and adding to validation

MaxAlex:
Previous addition to the Masabi pipeline used an incorrect URL variable, which is fixed here.

To prevent this kind of difficulty in the future, URL env variables are also added to the validation check on Odin start.


ealexa05 (approved):
This looks good. I think it's appropriate to keep the URLs in private, but it's probably not strictly necessary

PR #124, opened 2026-04-06
Chore: Increase Masabi data row limit

MaxAlex:
Towards end-to-end testing of the Masabi pipeline (https://app.asana.com/1/15492006741476/project/1212830297996795/task/1213666250518227?focus=true) update the row limit to a sensible production figure.


ealexa05 (approved):
This seems like a good safe starting point, but we should log a todo to find out our expected volume of masabi data on the prod side

edit: it looks like there are ~200-250k commuter rail trips every weekday, which means for a table with 1 row per ride this would need to run 25 times a day. Which could be fine, but worth thinking through

PR #125, opened 2026-04-06
Feat: Add status-based rescheduling behavior for Masabi

MaxAlex:
Have the Masabi pipeline reschedule itself in a shorter time if there is still non-ingested data waiting.

### DIFF
--- a/src/odin/ingestion/masabi/masabi_archive.py
+++ b/src/odin/ingestion/masabi/masabi_archive.py
@@ -11,7 +11,7 @@ import polars as pl
 import pyarrow.parquet as pq
 
 from odin.utils.logger import ProcessLog
-from odin.job import NEXT_RUN_DEFAULT, OdinJob, job_proc_schedule
+from odin.job import OdinJob, job_proc_schedule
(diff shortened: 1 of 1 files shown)


ealexa05 (changes requested):
I think it would be good to stick with the constants that the other jobs use for consistency:
```
NEXT_RUN_DEFAULT = 60 * 60 * 4  # 4 hours
NEXT_RUN_IMMEDIATE = 60 * 5  # 5 minutes
NEXT_RUN_LONG = 60 * 60 * 12  # 12 hours
```
(and eventually so that we can control these more globally, though they are currently all defined anew each script)

src/odin/ingestion/masabi/masabi_archive.py
```diff
@@ -53,7 +53,9 @@
 API_MIN_REQUEST_INTERVAL_S: float = 1.0
 
 # Time before rescheduling the job if there's remaining data.
```
ealexa05: Nit, but can you change this comment too?

ealexa05 (approved):
LGTM!

PR #126, opened 2026-04-06
feat: Addressing bug in load_cdc_records, adding test suite

ealexa05:
### Background
`ods_fact.py` has historically maintained 3 separate accumulators (`insert_df`, `update_df`, `delete_df`) for cdc operations that are incrementally built via `cdc_to_fact()`. This interleaved approach made it difficult to reason about operation conflicts, which, combined with a complete lack of unit tests, led to a bug where keys that were updated and subsequently deleted in an update batch were getting reinserted after the delete, which we saw impact the data in the unsettled transactions fact tables.

### Solution
This PR does 2 things to address this and prevent similar errors from being reintroduced in the future:

1. introduces a simplified structure for `load_cdc_records()` by accumulating all CDC records as raw data first, then resolving each key to its final operation (I/U/D) in one pass
2. introduces a test suite (in collaboration with @MaxAlex)

Here is the order of operations for the new structure (per the new docstring):

```
Order of operations:
  1. Read the current fact table and its max header__change_seq.
  2. Pull CDC records (I/U/D) from history with seq > max_fact_seq.
     Repeat in a loop to accumulate enough work.
  3. For every key touched by CDC, determine the FINAL operation:
     sort all CDC records by header__change_seq descending, deduplicate
     by key keeping the latest. The latest oper wins.
  4. Build a single "new_rows" dataframe:
     - For keys whose final op is I: use the I record directly.
     - For keys whose final op is U: fetch the existing fact row, apply
       the sparse column updates, produce an updated row.  If no fact
       row exists (I→U in same batch), use the I record as the base.
     - For keys whose final op is D: no new row (just drop the old one).
  5. Collect odin_index values of all fact rows being replaced or deleted
     (any key in the CDC batch that already exists in fact_ds).
  6. Write: filter old dataset to exclude touched rows, union with new_rows,
     upload to S3.
```

Important to note: since updates are sparse (not every column need be updated at once), when the final operation is U, all updates are applied in sequence. If a key is inserted and updated within a single batch, the insert acts as a base for the subsequent updates. When I or D operations are last, only the last operation for this key need be applied.

### Post-deploy operations
- [ ] Audit tables to determine which were affected
- [ ] Trigger a full snapshot reload for each affected table via migration
- [ ] Update downstream users on status

---
  - https://app.asana.com/0/0/1214063772751888

### DIFF
--- a/src/odin/generate/cubic/ods_fact.py
+++ b/src/odin/generate/cubic/ods_fact.py
@@ -393,50 +393,45 @@ class CubicODSFact(OdinJob):
         # then fetch existing fact rows and apply the updates.
         update_keys = resolved_updates.select(keys)
         if update_keys.height > 0:
-            # Build per-column update values from all U records matching update keys
-            u_records = cdc_df.cast(mod_cast).filter(pl.col("header__change_oper").eq("U"))
-            update_values = update_keys.clone()
-            for col in u_records.columns:
-                if col in keys:
-                    continue
-                col_df = (
-                    u_records.filter(pl.col(col).is_not_null())
-                    .sort(by="header__change_seq", descending=True)
-                    .unique(keys, keep="first")
-                    .select(keys + [col])
-                )
-                if col_df.height == 0:
-                    continue
-                if col in update_values.columns:
-                    update_values = update_values.pipe(pl_pipe_update, col_df, keys)
-                else:
-                    update_values = update_values.join(
-                        col_df, on=keys, how="left", nulls_equal=True, coalesce=True
-                    )
+            # Build per-key sparse update values: for each non-key column, take
+            # the latest non-null value across all U records for that key.
+            # Pre-sorting by seq desc means group_by().agg(drop_nulls().first())
+            # picks the highest-seq non-null value per column per key.
+            u_records = (
+                cdc_df.cast(mod_cast)
+                .filter(pl.col("header__change_oper").eq("U"))
+                .join(update_keys, on=keys, how="semi", nulls_equal=True)
+                .sort(by="header__change_seq", descending=True)
+            )
+            non_key_cols = [c for c in u_records.columns if c not in keys]
+            update_values = u_records.group_by(keys).agg(
+                [pl.col(c).drop_nulls().first() for c in non_key_cols]
+            )
 
-            # Fetch existing fact rows for these keys and apply updates
+            # Fetch existing fact rows for these keys
             existing_rows = ds_batched_join(
                 fact_ds, update_values.cast(orig_cast), keys, self.batch_size
             )
 
             # I→U in same batch: key was inserted and then updated, so no
             # existing fact row exists. Fall back to the I record as the base.
-            if existing_rows.height < update_keys.height:
-                found_keys = existing_rows.select(keys).cast(mod_cast)
-                missing_keys = update_keys.join(found_keys, on=keys, how="anti", nulls_equal=True)
-                if missing_keys.height > 0:
-                    insert_base = (
-                        cdc_df.cast(mod_cast)
-                        .filter(pl.col("header__change_oper").eq("I"))
-                        .sort(by="header__change_seq", descending=True)
-                        .unique(keys, keep="first")
-                        .join(missing_keys, on=keys, how="inner", nulls_equal=True)
-                    )
-                    if insert_base.height > 0:
-                        existing_rows = pl.concat(
-                            [existing_rows.cast(mod_cast), insert_base],
-                            how="diagonal",
-                        ).cast(orig_cast)
+            # Note: resolved_inserts cannot be used here because it only contains
+            # keys whose FINAL operation is I. For I→U keys the final op is U,
+            # so the I record only exists in the raw cdc_df.
+            found_keys = existing_rows.select(keys).cast(mod_cast)
+            missing_keys = update_keys.join(found_keys, on=keys, how="anti", nulls_equal=True)
+            if missing_keys.height > 0:
+                insert_base = (
+                    cdc_df.cast(mod_cast)
+                    .filter(pl.col("header__change_oper").eq("I"))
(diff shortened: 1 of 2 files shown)


src/odin/generate/cubic/ods_fact.py
```diff
@@ -363,243 +297,267 @@ def load_new_snapshot(self) -> None:
...
+                        .unique(keys, keep="first")
+                        .join(missing_keys, on=keys, how="inner", nulls_equal=True)
+                    )
+                    if insert_base.height > 0:
+                        existing_rows = pl.concat(
+                            [existing_rows.cast(mod_cast), insert_base],
+                            how="diagonal",
+                        ).cast(orig_cast)
+
+            if existing_rows.height > 0:
```
MaxAlex, line 436: Code style opinion: some of these conditionals are best left out, for the sake of readability and consistency; e.g., here there isn't a special case where `existing_rows` is non-empty which needs particular handling, rather there's an operation that will always be done to the contents of `existing_rows`, and it's a no-op if `existing_rows` is empty. Letting the same code run on empty and non-empty code is simpler to think about and gives fewer opportunities for edge-case-specific bugs to come up.

I think similar arguments can be made for the conditionals on 397, 426, 437, 460 and 466, assuming the respective code behaves as expected on empty data. (429 gets a pass because it looks like that does an operation on `cdc_df` that would be expensive even if `missing_keys` is empty.)

In this case, also, if you take out this  conditional (and the one on 397) you can also take out the separate definition of `new_update_rows` on 396.

src/odin/generate/cubic/ods_fact.py
```diff
@@ -363,243 +297,267 @@ def load_new_snapshot(self) -> None:
...
-                history_min_seq=str(hist_min),
-                history_max_seq=str(hist_max),
-            )
-        else:
-            cdc_log.add_metadata(
-                cdc_seq_min=str(cdc_df.get_column("header__change_seq").min()),
-                cdc_seq_max=str(cdc_df.get_column("header__change_seq").max()),
+        all_cdc_frames: list[pl.DataFrame] = []
+        current_min_seq = max_fact_seq
+        max_load_records = 10_000
```
MaxAlex: Code nit: Constants should be moved to the top of the file and given an all-caps name.

src/odin/generate/cubic/ods_fact.py
```diff
@@ -363,243 +297,267 @@ def load_new_snapshot(self) -> None:
...
+        for _ in range(11):
```
MaxAlex, line 347: Similarly; do we know why this is 11?

src/odin/generate/cubic/ods_fact.py
```diff
@@ -363,243 +297,267 @@ def load_new_snapshot(self) -> None:
...
-        s3_objects = list_objects(f"s3://{self.s3_export}/", in_filter=".parquet")
+        # --- Step 1: Load current fact table state ---
         fact_ds = ds_from_path(f"s3://{self.s3_export}/")
-        fact_files = ds_files(fact_ds)
         initial_row_count = fact_ds.count_rows()
         _, max_fact_seq = ds_metadata_min_max(fact_ds, "header__change_seq")
+        _, max_odin_index = ds_metadata_min_max(fact_ds, "odin_index")
 
-        # Log initial fact table state
         init_log = ProcessLog(
```
MaxAlex: It's a pre-existing issue, but while we're fixing this function we can also fix how ProcessLog is being mis-used here. Should have one ProcessLog for the function, with further add_metadata(), .complete() and .fail() as appropriate.

src/odin/generate/cubic/ods_fact.py
```diff
@@ -363,243 +297,267 @@ def load_new_snapshot(self) -> None:
...
+             - For keys whose final op is D: no new row (just drop the old one).
+          5. Collect odin_index values of all fact rows being replaced or deleted
+             (any key in the CDC batch that already exists in fact_ds).
+          6. Write: filter old dataset to exclude touched rows, union with new_rows,
+             upload to S3.
         """
-        # Load fact dataset and get current max sequence
-        s3_objects = list_objects(f"s3://{self.s3_export}/", in_filter=".parquet")
+        # --- Step 1: Load current fact table state ---
         fact_ds = ds_from_path(f"s3://{self.s3_export}/")
```
MaxAlex: Nit: should use the S3 directory path utility function from earlier.

MaxAlex (changes requested):

src/odin/generate/cubic/ods_fact.py
ealexa05: Good call, I consolidated the existing log functions into one

src/odin/generate/cubic/ods_fact.py
ealexa05: Fixed this to use your S3 directory path utility function

src/odin/generate/cubic/ods_fact.py
ealexa05: This is pretty arbitrary, but seemed to be working ok before

src/odin/generate/cubic/ods_fact.py
ealexa05: Made a `MAX_LOAD_RECORDS` constant (which `max_load_records` loads, as it is not a constant)

src/odin/generate/cubic/ods_fact.py
ealexa05: I think I've simplified this as much as makes sense for now. I took out the separate definition of `new_update_rows`, as the next definition is no longer gated by `if existing_rows.height > 0:`. I did keep `if existing_rows.height > 0:` though just so casting `odin_year` only runs when there is existing data. For the other places, it looks like it saves time to keep the special cases for now

MaxAlex (approved):

PR #128, opened 2026-04-09
Chore: Move view queries out to individual files

MaxAlex:
cubic_reports_sql.py was at 1659 lines and expanding rapidly, so it seems like a good time to separate view queries out into their own files. This will also make adding new views easier to manage in the future, since we'll no longer have to edit duck_db.py.

To validate that the view structure is unchanged after this goes in, I've copied out the current view definitions from the current dev and prod index files. The output of my local test (`fares_data_repository_no_env.db`) is consistent; we can additionally check that the result from running on dev is identical to prod before fully deploying.

```
prod_views = dict(prod_con.execute("""
    SELECT view_name, sql
    FROM duckdb_views()
    WHERE schema_name = 'cubic_reports'
""").fetchall())

with open('prod_view_dump_040926.json', 'w') as out:
    json.dump(prod_views, out)
```

[prod_view_dump_040926.json](https://github.com/user-attachments/files/26613856/prod_view_dump_040926.json)
[dev_view_dump_040926.json](https://github.com/user-attachments/files/26613857/dev_view_dump_040926.json)


ealexa05 (approved):
This looks good to me! Should make maintaining this all a lot easier

PR #134, opened 2026-04-17
Chore: Enlarge batch size to hopefully unstick some tables

MaxAlex:
Quick fix for https://app.asana.com/1/15492006741476/project/1212830297996795/task/1213450668245131?focus=true

Will cause CDC ingestion to use a larger batch size across all tables; this should be reverted after the immediate issue with edw_farerev_recovery_txn (and possibly other tables) is resolved, to avoid unnecessary OOM crashes.


ealexa05 (approved):
Approving, but let's hold off a minute on the deploy. I can definitely see this causing OOM, so we should do our easy regeneration tasks first and be ready to revert this quickly (agree that it should be temporary)

PR #140, opened 2026-04-24
fix: remove CERT_NONE flag from masabi urllib pool

MaxAlex:
Asana ticket: https://app.asana.com/1/15492006741476/project/1212830297996795/task/1213877775425747?focus=true

Was originally put in to avoid a certificate error encountered during local testing. I'm fairly certain it was zscaler-related, so it shouldn't ever effect dev or prod, but it's worth testing that out sooner than later.


ealexa05 (commented):
Looks good! Let's keep it dev over the weekend

ealexa05 (approved):

PR #147, opened 2026-05-01
Add new tables requested by Hui, and change ingestion start point

MaxAlex:
Asana ticket: https://app.asana.com/1/15492006741476/project/1203930661880158/task/1214070415909699?focus=true

Updates pipeline to add new tables and change the start date to the beginning of 2020.

The original tables won't immediately update from the new start in dev, but everything will be starting new from that time when we move to prod.


ealexa05 (approved):
Looks good to me! Will be interested to see how much data this is / how long ingestion takes

PR #148, opened 2026-05-06
Use different timestamp key for tables missing serverTimestamp

MaxAlex:
Earlier we added a new set of tables to the Masabi pipeline. It turns out that a bunch of these, all the tables in the 'view' schema, don't have a serverTimestamp column. They do have a 'latestIngestTimestamp' column which appears to be similar; my guess is, on Masabi's side, these are a separate database of constructed report views, and latestIngestTimestamp is tracking the serverTimestamp of the underlying data.

Also, a couple of further fixes that came up while doing this:

- The loop in fetch_and_write would hit an error on min() if it saw an empty page.
- One of the new tables, view.validators, has an 'hourOfDay' column that's string type rather than the expected numeric type.


ealexa05 (approved):
These changes all look good! Noting for posterity that `masabi_fetch_and_write` log will show `min_obs_ts=float("inf")`/ `max_obs_ts=-1` when it encounters an empty page

PR #150, opened 2026-05-06
Fix: Top level logging start-complete for all jobs

MaxAlex:
There are currently no particular log messages that consistently mark the beginning and end of jobs, making it difficult to determine how much time individual jobs or tables are taking. This adds a top-level log for every job that didn't previously have one, which is guaranteed to always emit a failed or complete message.


ealexa05 (approved):
Looks good to me!

PR #151, opened 2026-05-06
Fix: Add numeric override for validation.telemetry

MaxAlex:
validation.telemetry in the Masabi data has the same incorrectly typed hourOfDay column as a bunch of others.


ealexa05 (approved):
Approved. Have you checked through the other added tables to ensure you got all instances? hourOfDay seems to be incorrectly typed in the schema across the board

PR #153, opened 2026-05-11
Feat: Add PII exclusion to Masabi pipeline

MaxAlex:
Asana ticket: https://app.asana.com/1/15492006741476/project/1203930661880158/task/1213954203541908?focus=true

Adds the list of PII exclusion columns by table, and excludes these from retrieval and the table schema. Warnings on columns that are missing due to exclusion reasons should also be suppressed.

While we're at it, removes several tables for which there is currently no data, so we can't yet make a determination on the PII columns.

### DIFF
--- a/src/odin/ingestion/masabi/masabi_archive.py
+++ b/src/odin/ingestion/masabi/masabi_archive.py
@@ -61,8 +61,6 @@ NEXT_RUN_LONG = 60 * 60 * 12  # 12 hours
 MASABI_START_TIMESTAMP_MS: int = 1_577_836_800_000
 
 TABLES = [
-    "data.external_rider_data",
-    "partner.rider_association_events",
     "retail.account_actions",
     "retail.activations",
     "retail.rider_entitlement_events",
@@ -72,7 +70,6 @@ TABLES = [
     "validation.telemetry",
     "view.hub_search_account",
     "view.hub_search_guest_account",
-    "view.hub_search_token",
     "view.hub_search_vendor_sale",
     "view.validators",
 ]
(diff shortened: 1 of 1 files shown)


src/odin/ingestion/masabi/masabi_archive.py
```diff
@@ -439,7 +492,7 @@ def _make_request(
         self,
         pool: urllib3.PoolManager,
         url: str,
-        fields: dict[str, str],
+        fields: list[tuple[str, str]],
```
ealexa05, line 495: Why was this changed to a list of tuples?

ealexa05: i.e. `fields.append(("exclude", exc_col))`, rather than `fields['exc_col'].append(exc_col)`

ealexa05: It just seems weird data structure-wise to have `orderBy`, `size`, etc. be tuples

ealexa05 (approved):

src/odin/ingestion/masabi/masabi_archive.py
MaxAlex: The Masabi API expects multiple 'exclude' rows to be in the form of multiple 'exclude=' arguments, which we can't do if we're passing the arguments as a dict. This otherwise logically equivalent for urllib3. (Agreed it's less elegant, though.)

ealexa05: Oh that's annoying on Masabi's part, but ok that makes sense!

PR #155, opened 2026-05-12
Fix: Disable Masabi view tables

MaxAlex:


ealexa05 (approved):
Looks good! I think punting on views is the right call

PR #156, opened 2026-05-13
feat: Splitting jobs and tables across new instances

ealexa05:
Adding ability to split tables and job types across the instances soon to be created in https://github.com/mbta/terraform_modules/pull/790, mediated by a new function `get_odin_instance()`.

When running in aws, jobs are set to run based on their specific instance of Odin ("alpha" or "beta"), as are specific tables. When running locally, set the environment variable `ODIN_INSTANCE` to one of these to run the jobs and tables associated with that instance.

As before, logic dictating which jobs run lives in [src/odin/run.py](src/odin/run.py). This PR transitions away from using toml files to determine which jobs to run to simplify coordination when jobs and tables are each getting split, and to remove the need for a devops PR cycle every time we move a table.

Once this PR is implemented, it will be safe to remove the job config tomls from devops (e.g. [odin-prod.toml](https://github.com/mbta/devops/blob/d9e97a8cbbc1cafe25e45b054d3616faf5a8e0a0/terraform/aws-tid-main-prod/json_config/odin-prod.toml))

### Tables
`ArchiveCubicQlikTable`, `CubicODSFact`, `ArchiveAFCAPI`, and `ArchiveMasabi` consult table lists, and this PR introduces lists linked to each (e.g. `CUBIC_ODS_TABLES_ALPHA`). The schedulers for of these jobs run on each instance, but only schedule jobs when tables are specifically slated for that instance.

Each table should be fully owned by exactly one instance, and setting it up this way ensures that all jobs consulting a given table list (e.g. `CubicODSFact` and `ArchiveAFCAPI`) stay in sync. Since jobs do not run in parallel on a given instance, we can't end up in a situation where `CubicODSFact` and `ArchiveAFCAPI` are running simultaneously for a given table, which could have unintended consequences.

### Jobs
`RestrictedAFC` will be changing soon, works differently, and only contains one table ("v_media"), so it is just limited so that it only runs on the alpha instance.

`DataDictionary` is lightweight and shouldn't be running simultaneously on multiple instances, so it is set here to run only in the alpha instance.

### Beta instance testing / division of labor
After this PR is merged, the alpha instance will continue to do exactly what odin-prod was doing. The beta instance will start syncing "EDW.TRANSACTION_HISTORY", which was recently disabled and is currently unessential, but will be a good test of capacity. Once we have verified this is working, we can move over the high priority tables to ensure they are unblocked, and remove "EDW.TRANSACTION_HISTORY"

### check_for_parallel_tasks
Each instance will have its own `ECS_TASK_GROUP` (defined as `family:odin-${each.key}-${var.environment}`), so these shouldn't interfere with each other

### Data dictionary
This is based on `AWS_ENV = os.getenv("ECS_TASK_GROUP", "no_env").split("-")[-1]`, which will result in `${var.environment}` as it did before, so these will remain in their current S3 locations.

### Migrations
Migrations are based on `task_name = task_name.replace("family:", "")`, so this will change locations. New migrations for prod will start at [src/odin/migrate/migrations/odin-alpha-prod/0001.py](src/odin/migrate/migrations/odin-alpha-prod/0001.py) and be tracked in s3://mbta-ctd-dataplatform-archive/odin/migrations/odin-alpha-prod/, and dev in odin-alpha-dev.

### GH deploy workflows
Deploy workflows now run for "all" instances (i.e. both "alpha" and "beta")*, and individual instances can be paused. Most practically, if you want to make sure your migration runs on alpha before the beta instance starts working on its jobs, you can now pause just beta to allow alpha to finish its migration work

* I have put in code to deploy for individual instances if we need that in the future (e.g. for flexibility if you don't want to interrupt a long-running or critical job on one instance), but it is currently not accessible because "alpha" and "beta" are not enabled as options for manual deploy, because it is probably a bad idea to do this

### Logging
Both instances will dump logging to `index="odin-prod"` (or `dev`). In order to disambiguate the logs, this PR adds `odin_instance` and `ecs_task_group` to the default logging so that you can filter results to just the instance you are looking at.

---
[Asana](https://app.asana.com/1/15492006741476/project/1214070495342435/task/1214970995873228?focus=true)


.github/workflows/deploy_base.yaml
```diff
@@ -4,9 +4,14 @@ on:
   workflow_call:
     inputs:
       environment:
-        description: One of 'prod', 'staging', or 'dev'
+        description: "One of 'prod', 'staging', or 'dev'"
         required: true
         type: string
+      odin-instance:
```
MaxAlex, line 10: I'm not sure deploying only one instance is a capability we even want to have; things could get very confusing if the instance code is out of sync.

src/odin/generate/cubic/ods_fact.py
```diff
@@ -35,7 +35,9 @@
 from odin.utils.aws.s3 import s3_folder
 from odin.ingestion.qlik.dfm import dfm_from_s3
 from odin.ingestion.qlik.dfm import QlikDFM
-from odin.ingestion.qlik.tables import CUBIC_ODS_TABLES
+from odin.ingestion.qlik.tables import CUBIC_ODS_TABLES_ALPHA
```
MaxAlex: As before, the alpha/beta table list split should be handled solely in tables.py.

src/odin/ingestion/afc/afc_archive.py
```diff
@@ -87,6 +88,10 @@ class APIEmptyResponseError(Exception):
     "v_products",
 ]
 
+API_TABLES_BETA: list[str] = []
+
+API_TABLES = API_TABLES_ALPHA + API_TABLES_BETA
```
MaxAlex: It's alright to leave AFC instance table list logic in afc_archive.py, but it would be a lot cleaner to separate it from the processing logic. API_TABLES should be set to either API_TABLES_ALPHA or API_TABLES_BETA depending on the instance, and the rest of the code should be unchanged.

src/odin/ingestion/masabi/masabi_archive.py
```diff
@@ -794,6 +799,8 @@ def schedule_masabi_archive(schedule: sched.scheduler) -> None:
 
     :param schedule: application scheduler
     """
-    for table in TABLES:
+    instance = get_odin_instance()
+    instance_tables = TABLES_ALPHA if instance == "alpha" else TABLES_BETA
```
MaxAlex: Ditto from the afc_archive.py comment.

src/odin/ingestion/qlik/cubic_archive.py
```diff
@@ -372,7 +374,9 @@ def schedule_cubic_archive_qlik(schedule: sched.scheduler) -> None:
 
     :param schedule: application scheduler
     """
-    for table in CUBIC_ODS_TABLES:
+    instance = get_odin_instance()
+    instance_tables = CUBIC_ODS_TABLES_ALPHA if instance == "alpha" else CUBIC_ODS_TABLES_BETA
```
MaxAlex: Ditto from ditto

.github/workflows/deploy_base.yaml
ealexa05: Updated to new strategy:

> Deploy workflows now run for "all" instances (i.e. both "alpha" and "beta")*, and individual instances can be paused. Most practically, if you want to make sure your migration runs on alpha before the beta instance starts working on its jobs, you can now pause just beta to allow alpha to finish its migration work with Change Task Count

src/odin/generate/cubic/ods_fact.py
ealexa05: Done

src/odin/ingestion/afc/afc_archive.py
ealexa05: Done

src/odin/ingestion/masabi/masabi_archive.py
ealexa05: Done

src/odin/ingestion/qlik/cubic_archive.py
ealexa05: Done

MaxAlex (approved):

PR #158, opened 2026-05-15
Add Masabi restricted schema data

MaxAlex:
Asana ticket: https://app.asana.com/1/15492006741476/project/1212830297996795/task/1214710854658380?focus=true

Adds a mode to the ArchiveMasabi writer that writes data with the restricted-allowed PII columns to a separate restricted masabi data path, and a new ViewBuilder instance that brings that data into the catalog.

This should go in with the corresponding devops PR: https://github.com/mbta/devops/pull/4370


ealexa05 (approved):
Looks good to me!

PR #159, opened 2026-05-20
Fixes to restricted schema logic

MaxAlex:
- Run restricted-mode ingestion only for tables that have restricted-allowed columns, not all tables that have PII
- "location" was missing from the PII list for retail.ticket_purchases, even though it was on the restricted-allowed list


ealexa05 (approved):
Good catch!

PR #168, opened 2026-06-04
chore: Increase Masabi maximum rows per run

MaxAlex:
Increasing the row limit so that we'll catch up faster.


ealexa05 (approved):
I think this will help, approved

PR #169, opened 2026-06-04
Temporarily restricted schema updates (and shorter immediate re-run)

MaxAlex:
Disabling the restricted schema updates so that the main tables catch up faster.

Also shortening NEXT_RUN_IMMEDIATE; we may run into a case where there's only one or two of the largest tables left catching up, which would mean that Odin-beta wouldn't saturate with a five-minute delay.

---
  - https://app.asana.com/0/0/1215397953195070

### DIFF
--- a/src/odin/ingestion/masabi/masabi_archive.py
+++ b/src/odin/ingestion/masabi/masabi_archive.py
@@ -55,7 +55,7 @@ API_MIN_REQUEST_INTERVAL_S: float = 1.0
 
 # Rescheduling time intervals
 NEXT_RUN_DEFAULT = 60 * 60 * 4  # 4 hours
-NEXT_RUN_IMMEDIATE = 60 * 5  # 5 minutes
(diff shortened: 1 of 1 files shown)


ealexa05 (approved):
Approving, should give us time to implement the restricted tables in a more efficient way if that's something we need (although, the best way probably doesn't work retroactively)

src/odin/ingestion/masabi/masabi_archive.py
```diff
@@ -55,7 +55,7 @@
 
 # Rescheduling time intervals
 NEXT_RUN_DEFAULT = 60 * 60 * 4  # 4 hours
-NEXT_RUN_IMMEDIATE = 60 * 5  # 5 minutes
+NEXT_RUN_IMMEDIATE = 60  # 1 minute
```
ealexa05, line 58: Style nit: I'd keep as `60 * 1` since we intend to revert it

ealexa05 (approved):

PR #175, opened 2026-06-08
Fix masabi parquet rollover error

MaxAlex:
The old `pq_dataset_writer` function used for generating synced parquet files has an `export_folder` argument; this argument is undocumented, but it turns out that in addition to being the directory to which `pq_dataset_writer` writes its output, `pq_dataset_writer` must also contain the downloaded parquet files. This is because it will look in that directory to determine the correct filename to write (e.g., `table_002.parquet` if the highest-index table downloaded is `table_002.parquet`, etc.) If the provided `export_directory` is empty, it will silently default to `table_001.parquet`; if `export_directory` does not exist, it will **silently create that directory** and then read it, seeing no files, and likewise default to table_001.parquet.

When a table had previously collected enough data that it was split between two files, e.g. `table_001.parquet` and `table_002.parquet`, the pipeline would download `table_002.parquet`, append new data to it, and then, if it sees an empty `export_folder`, write out the result into a new table_001.parquet. This is then uploaded, overwriting table_001.parquet with data from table_002.parquet and later. All subsequent updates will follow the same behavior, overwriting table_001.parquet with table_002.parquet, plus the most recently downloaded data.

For non-Masabi uses, there are different path constructions for the parquet local download target and for `export_folder`, but these wind up being the same in each case. The difference in Masabi's case was due to the addition of the `s3_folder` function to the `ArchiveMasabi.export_folder`, which we started using after the earlier Cubic bug which was due to inconsistent S3 path handling; this added the "s3://" prefix to the path, changing the path provided to `export_folder` and triggering the behavior described above.

As a minimal fix, this PR changes the Masabi pipeline so that the directory sent to `export_folder` is guaranteed to contain the downloaded parquet file, unless there is no pre-existing data.

(This PR includes some changes for validating the parquet rollover behavior which will be removed prior to merging.)

### DIFF
--- a/src/odin/ingestion/masabi/masabi_archive.py
+++ b/src/odin/ingestion/masabi/masabi_archive.py
@@ -55,7 +55,7 @@ API_MIN_REQUEST_INTERVAL_S: float = 1.0
 
 # Rescheduling time intervals
 NEXT_RUN_DEFAULT = 60 * 60 * 4  # 4 hours
-NEXT_RUN_IMMEDIATE = 1  # 1 minute  ## TESTING
+NEXT_RUN_IMMEDIATE = 1  # 1 minute  ### TESTING
 NEXT_RUN_LONG = 60 * 60 * 12  # 12 hours
 
 # Exclusive lower bound for the initial historical backfill: 2020-01-01 00:00:00 UTC (ms).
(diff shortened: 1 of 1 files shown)


ealexa05 (approved):
Approving, though adding the context from the intro to the pq_dataset_writer could be nice:

> The old pq_dataset_writer function used for generating synced parquet files has an export_folder argument; this argument is undocumented, but it turns out that in addition to being the directory to which pq_dataset_writer writes its output, pq_dataset_writer must also contain the downloaded parquet files. This is because it will look in that directory to determine the correct filename to write (e.g., table_002.parquet if the highest-index table downloaded is table_002.parquet, etc.) If the provided export_directory is empty, it will silently default to table_001.parquet; if export_directory does not exist, it will silently create that directory and then read it, seeing no files, and likewise default to table_001.parquet.

PR #179, opened 2026-06-10
chore: materialize comp views, move to beta

MaxAlex:
In response to performance issues with the comp A/B views, this adds those views as materialized views*. Local testing has shown this to run <30 min per table, but this could still be a significant time drain on Odin; depending on how performance looks on the ECS, we may want to optimize further.

(*duckdb doesn't have materialized views, but the pipeline fakes it by building a new parquet and defining a view onto that.)


ealexa05 (approved):
Haven't run the views to test, but otherwise looks good to me!

PR #182, opened 2026-06-10
fix: workload tuning additions for mat view performance

MaxAlex:
Trying to materialize the COMP B views currently fails:

```
7959f6122a6 ERROR uuid=3e99cd25-92e2-4cbf-a101-e7f43da2247d, File "/usr/local/lib/python3.13/site-packages/odin/generate/data_dictionary/duck_db.py", line 129, in create_fares_db
87959f6122a6 ERROR uuid=3e99cd25-92e2-4cbf-a101-e7f43da2247d, con.execute(mat_view_query)
87959f6122a6 ERROR uuid=3e99cd25-92e2-4cbf-a101-e7f43da2247d, ~~~~~~~~~~~^^^^^^^^^^^^^^^^
87959f6122a6 ERROR uuid=3e99cd25-92e2-4cbf-a101-e7f43da2247d, _duckdb.OutOfMemoryException: Out of Memory Error: could not allocate block of size 256.0 KiB (12.7 GiB/12.7 GiB used)
87959f6122a6 Possible solutions:
87959f6122a6 * Reducing the number of threads (SET threads=X)
87959f6122a6 * Disabling insertion-order preservation (SET preserve_insertion_order=false)
87959f6122a6 * Increasing the memory limit (SET memory_limit='...GB')
87959f6122a6 See also https://duckdb.org/docs/stable/guides/performance/how_to_tune_workloads
87959f6122a6 INFO uuid=3e99cd25-92e2-4cbf-a101-e7f43da2247d, parent=odin, process=create_report_mat_views, odin_instance=beta, ecs_task_group=family:odin-beta-prod, process_id=375, status=failed, disk_free_mb=95682, sys_mem_free_pct=85, proc_mem_used_mb=10253, duration=418.02, error_type=OutOfMemoryException, view_name=mat_comp_b_txn_a
```

This PR adds several of the workload tuning measures noted in the link; hopefully this will get the view materialization through.

NB: Limiting the thread count may slow down everything else that duck_db.py does; we can re-evaluate if the view materialization works and we keep it around.

PR #183, opened 2026-06-10
fix: further duckdb pragmas, and isolated edw_farerev_payg_trip_txn s…

MaxAlex:
Following https://github.com/mbta/odin/pull/182 ; COMP B materialized views are still triggering duckdb's OOM safeguards on Odin prod. This also adds edw_farerev_payg_trip_txn.sql, which is the major sub-expression in both COMP B queries, so if that works we can simply redefine the COMP B views to reference that.


ealexa05 (approved):
These all seem like safe optimizations (worst it could do is crash which it is doing already), so I'm good to go ahead and test

PR #184, opened 2026-06-10
Fix even more comp b duckdb tuning

MaxAlex:
One more slight change to try and get COMP B view materialization to work; spill data path that's guaranteed to already exist.


ealexa05 (approved):
approved pending `ruff format --diff .`

PR #187, opened 2026-06-12
Add 'V2' comp b views leveraging restructured+materialized farerev_pa…

MaxAlex:
Attempting to materialize the COMP B views in order to address performance issues has consistently ran into memory issues. as did splitting out the edw_farerev_payg_trip_txn subexpression as its own materialized view. This restructures edw_farerev_payg_trip_txn in order to reduce memory use, while keeping the result the same. This also adds prototype 'v2' versions of the final COMP B view which use the new materialized edw_farerev_payg_trip_txn.


ealexa05 (approved):
Approved!

PR #188, opened 2026-06-15
Third attempt at tuning duckdb for comp b views

MaxAlex:
Third attempt at tuning duckdb performance to get the COMP B views through without memory errors; removes the previous manual memory_limit, and lowers thread count to 1 for materialized view creation.


ealexa05 (approved):
🤞

PR #190, opened 2026-06-18
Fix S3 path handling bug in AFC update code

MaxAlex:
Asana (at least mostly related): https://app.asana.com/1/15492006741476/project/1213716198445326/task/1215684729287628?focus=true

It turned out that v_entitlements was consistently missing all data except for the most recent update. This was due to improper S3 path handling in the old list_objects() utility function leading to a collision with v_entitlements_full, loading in the wrong file, and incorrect syncing left only the new data being uploaded (which simultaneously overwrote the old, complete v_entitlements parquet.)

This fix is intentionally minimal to avoid running into other issues with the parquet utility code. (E.g., the naive fix, changing the definition of AFCArchive.export_folder to use s3_folder directly, would have caused the same issue that deleted five years' worth of Masabi data last week.)


ealexa05 (approved):
Looks good! Now for the rest of the S3 paths in Odin...

PR #191, opened 2026-06-19
feat: lightweight monitoring for AFC pipeline

ealexa05:
This is part of a larger push to implement monitoring for each of the Odin sources (Cubic ODS, AFC, Masabi, Spare), and is being expedited to allow us to check for regressions in https://github.com/mbta/odin/pull/190. This specifically covers metrics which are lightweight to compute and useful to tie to individual runs.

Before and after each run of `sync_parquet()`, this code saves an `AFCParquetSnapshot` object, which contains:
- `object_count`
- `total_size_bytes`
- `total_rows`
- `min_job_id`
- `max_job_id`

which can all be collected efficiently from S3 using parquet metadata across files in partition using `list_objects` and `ds_metadata_min_max`. Checks of S3 data use `s3_folder`, to avoid the problem to be addressed in https://github.com/mbta/odin/pull/190 and to check only the specific table being run.

Because AFC data is strictly additive in batches marked by jobId, none of `object_count`, `total_size_bytes`, `total_rows`, or `min_job_id` should decrease. This code emits a `afc_api_inline_health_regression` log with `status=failed` if `total_size_bytes`, `total_rows`, `object_count` or `max_job_id` decreases after writing data to S3 (or if `min_job_id` increases), and logs each value pair as a batch for debugging purposes. Logged failures do not interrupt operation of pipeline, but will generate Splunk alerts downstream.

Asana: [Add lightweight monitoring for AFC pipeline](https://app.asana.com/1/15492006741476/project/1213716198445326/task/1215865675975562?focus=true)

### DIFF
--- a/src/odin/ingestion/afc/afc_archive.py
+++ b/src/odin/ingestion/afc/afc_archive.py
@@ -22,6 +22,7 @@ from odin.utils.aws.s3 import list_objects
 from odin.utils.aws.s3 import download_object
 from odin.utils.aws.s3 import upload_file
 from odin.utils.aws.s3 import delete_objects
+from odin.utils.aws.s3 import S3Object
 from odin.ingestion.afc.afc_tables import API_TABLES_INSTANCE
 from odin.ingestion.afc.afc_tables import _ODIN_INSTANCE
 from odin.utils.parquet import ds_metadata_min_max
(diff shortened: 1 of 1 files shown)


src/odin/ingestion/afc/afc_archive.py
```diff
@@ -211,6 +222,132 @@ def make_request(
...
+        """Collect near-zero-cost S3 parquet metrics for table health checks."""
+        snapshot: AFCParquetSnapshot = {
+            "object_count": 0,
+            "total_size_bytes": 0,
+            "total_rows": 0,
+            "min_job_id": None,
+            "max_job_id": None,
+        }
+
+        s3_objects = s3_objects or list_objects(s3_folder(self.export_folder), in_filter=".parquet")
```
MaxAlex, line 247: I would like it better if these were in a separate class in a separate file, just for code separation; there's a nice logical distinction between the update logic and the health check logic.

MaxAlex (approved):
As per comment; I think the new code would be better split off into a separate file.

src/odin/ingestion/afc/afc_archive.py
ealexa05: Noted! I will try to take these functions out when I implement the next level of health check logic (as this function will also be good to have for the async check job)

PR #192, opened 2026-06-23
fix(regression tests): only check min_job_id if table is transactional; check size less strictly

ealexa05:
Follow-on to https://github.com/mbta/odin/pull/191, updating two regressions which are causing false positives:
- `min_job_id` seems to increase each time in static tables by design, so an increase in `min_job_id` is now only a regression for transactional tables
- `total_size_bytes` can decrease slightly when data is added to the parquet file, presumably due to the compression process. Now only reductions in size greater than 1% of the original file are reported to logs as a regression (which, again, do not block any operations)

Asana: [Add lightweight monitoring for AFC pipeline](https://app.asana.com/1/15492006741476/project/1208949713596457/task/1215865675975562?focus=true)

### DIFF
--- a/src/odin/ingestion/afc/afc_archive.py
+++ b/src/odin/ingestion/afc/afc_archive.py
@@ -301,7 +301,8 @@ class ArchiveAFCAPI(OdinJob):
             )
 
         if (
-            pre_snapshot["min_job_id"] is not None
+            self.table_type == "transactional"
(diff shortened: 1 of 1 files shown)


MaxAlex (approved):
I suspect 0.99 is not nearly small enough of a constant there, but looks good!

PR #196, opened 2026-06-29
feat: Add OdinJob memory watch loop and log output.

MaxAlex:
Currently we report memory usage stats in ProcessLog output; however, this approach fails to capture real RAM usage as the check lands exactly when a process is completed, whereupon the memory it's using is potentially already freed. This adds a loop that polls current memory usage for the currently running OdinJob, and returns a log line with the highest observed RAM usage overall. The log reports OdinJob arguments so that we can trace the output to an individual job type, table, etc.

### DIFF
--- a/src/odin/job.py
+++ b/src/odin/job.py
@@ -1,6 +1,7 @@
 from abc import ABC
 from abc import abstractmethod
 from multiprocessing import get_context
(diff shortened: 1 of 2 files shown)


ealexa05 (approved):
Looks good! (for posterity, this was also tested locally and works well)

PR #197, opened 2026-06-29
fix: Remove unnecessary transit_day bounds on wo118 definition

MaxAlex:
Asana ticket: https://app.asana.com/1/15492006741476/project/1212830297996795/task/1216052827690186?focus=true

WO118 currently has a trailing WHERE clause that limits the results to a specified time range. In the linked Asana ticket, Will confirms that this was not meant to be included in the Odin copy of the query, so here we're removing it. (The missing data which is the topic of the ticket is apparently for separate reasons, though.)


ealexa05 (approved):
Approved!

PR #201, opened 2026-06-30
fix: comment out wa160's join to edw.member_dimension

MaxAlex:
Asana ticket: https://app.asana.com/1/15492006741476/project/1208949713596462/task/1215409134256424?focus=true

The join to edw_member_dimension by transit_account_id was non-unique, leading to unwanted duplication of rows in the output. The 'real' join in Cubic's verson of WA160 uses data that does not seem to be reflected in our copy currently. As requested by Will, this PR disables the join, preventing duplication, but losing the `external_ref` column.


ealexa05 (approved):
Looks good! Max confirmed the view output looks correct, and Will Keegan approved the SQL diff

PR #203, opened 2026-07-01
Delta Lake-based Cubic ODS fact table generation job

MaxAlex:
This adds a new OdinJob instance, CubicODSDelta, which sits parallel to the current CubicODSFact, reading the same history table and snapshot paths and writing fact tables (in delta lake’s parquet format) to a new fact table path: s3://mbta-ctd-dataplatform-dev-springboard/odin/data/cubic/ods_delta . This new path is given a new schema definition in duck_db.py, which is structured read the delta format via delta_scan, and so becomes a new schema `cubic_delta` available in the data repository.

Importantly, this new job makes no use of the parquet handling utility code in odin/src/utils/parquet.py. Fact table output is written entirely via the Python Delta Lake API; history table input is read via Python’s duckdb API. This is expected to make the job more reliable and faster than the existing CubicODSFact.


src/odin/generate/cubic/delta_ods.py
```diff
@@ -179,6 +187,31 @@ def _read_history(self) -> str:
         """Return the DuckDB read_parquet expression for this table's history."""
         return READ_HISTORY.format(glob=self.history_glob)
 
+    def _read_state(self) -> tuple[str, str]:
+        """
+        Return (snapshot, cdc_watermark) from the latest commit that recorded them.
+
+        A silver table with no recorded position (never built, or built by an
```
ealexa05, line 194: Want to think through a little more situations where we might be returning `("", INITIAL_WATERMARK)` erroneously. Is there any risk of setting `HISTORY_SCAN_LIMIT` too low, so that `STATE_SNAPSHOT_KEY in commit` returns `False`? Or having an empty value in `header__change_seq` triggering a reset even though we have data (like we were just guarding against for `SALE_TRANSACTION`)?

ealexa05 (approved):
Approving for testing in dev over the weekend, still want to think through this carefully. Exciting stuff!

src/odin/generate/cubic/delta_ods.py
ealexa05: Resolving so I can merge, but pls respond @MaxAlex

PR #204, opened 2026-07-01
fix: backfilling Masabi data (step 1)

ealexa05:
This PR starts backfilling Masabi data, which currently only goes back 1 year, to fill in the remainder of the requested 5 years (to 2021-01-01). Since this is a temporary measure and we want the backfilling processing to happen on odin-gamma, all Masabi tables running on odin-gamma will run in backfill mode (i.e. syncing 5 years of data to `MASABI_BACKFILL`).

Once this is up to date, a new migration will be created to swap in this data.


MaxAlex (approved):
I'd suggest but not require that the logic to set the backfill timestamp on gamma be in the definition of MASABI_START_TIMESTAMP_MS, rather than in the update logic itself. Possibly likewise for setting MASABI_DATA to the backfill directory. Otherwise, looks good.

PR #210, opened 2026-07-07
Several runtime optimizations for CubicODSDelta

MaxAlex:
Adds several performance optimizations to delta_ods.py:

- Increases MAX_MERGE_RECORDS to 500k, approximating the limit on fact_ods.py
- The 'more_pending' log flag is now based on whether we hit the merge records limit, rather than a separate scan of the fact tables
- When reading in CDC changes, instead of a single query for full data that uses LIMIT to restrict batch size, do two stages: first, a query just against the header__change_seq to find the upper and lower bound values that match the limit, and then a second query using those bounds. The previous version, which used `ORDER BY header__change_seq LIMIT {limit}` would require reading and sorting all header__change_seq values across all tables, rather than sorting only values that land within the interval.
- Restrict the merge predicate based on the odin month/year, so that the merge process is only touching files that need to be updated. This only works for updates which include the edw_inserted_dtm, but that looks like it's always the case (for tables that include that column.)

### DIFF
--- a/src/odin/generate/cubic/delta_ods.py
+++ b/src/odin/generate/cubic/delta_ods.py
@@ -505,10 +505,7 @@ class CubicODSDelta(OdinJob):
             return ""
         years_sql = ", ".join(str(y) for y in years)
         months_sql = ", ".join(str(m) for m in months)
-        return (
-            f' AND target."odin_year" IN ({years_sql})'
-            f' AND target."odin_month" IN ({months_sql})'
-        )
+        return f' AND target."odin_year" IN ({years_sql}) AND target."odin_month" IN ({months_sql})'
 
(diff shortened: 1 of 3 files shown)


src/odin/generate/cubic/delta_ods.py
```diff
@@ -81,7 +81,7 @@
 NEXT_RUN_LONG = 60 * 60 * 12  # 12 hours
 
 REBUILD_BATCH_SIZE = 10_000
-MAX_MERGE_RECORDS = 100_000
+MAX_MERGE_RECORDS = 500_000
```
ealexa05, line 84: Can confirm that it looks like we have plenty of memory headroom for this

ealexa05 (approved):
Confirming this as a safe change, will go ahead with dev performance test

PR #212, opened 2026-07-08
Set streamed_exec=False in delta merge

MaxAlex:
Possible optimization for the delta merge process, which seems to be the current bottleneck; we're not streaming in either case (we materialize `source` via `.to_arrow()`) and streamed_exec=True (the default) disables some methods deltalake uses to prune target files.

### DIFF
--- a/src/odin/generate/cubic/delta_ods.py
+++ b/src/odin/generate/cubic/delta_ods.py
@@ -552,6 +552,7 @@ class CubicODSDelta(OdinJob):
             error_on_type_mismatch=False,
             merge_schema=False,
             commit_properties=self._commit_state(watermark),
+            streamed_exec=False,
         )
         return (
             merger.when_matched_delete(predicate="source.header__change_oper = 'D'")


src/odin/generate/cubic/delta_ods.py
```diff
@@ -552,6 +552,7 @@ def _merge_apply(self, source: pl.DataFrame, keys: list[str], watermark: str) ->
             error_on_type_mismatch=False,
             merge_schema=False,
             commit_properties=self._commit_state(watermark),
+            streamed_exec=False,
```
ealexa05, line 555: This will increase our memory footprint, so I think you should maybe also bring the row limit back down with this?

ealexa05 (changes requested):

src/odin/generate/cubic/delta_ods.py
MaxAlex: I don't think it will, since we weren't really streaming the changes to begin with; it was working from a static source table already.

ealexa05: Ok, approving to test! We'll keep profiling memory usage

ealexa05 (approved):

PR #213, opened 2026-07-08
Additional fixes/improvements to delta fact job

MaxAlex:
Another set of changes to the delta fact table creation job:

- Fix to the handling of an edge case of successive updates: if a row is deleted and then re-inserted, nulls in the insert would have been interpreted as 'columns not being updated,' and would have reverted to the (potentially non-null) values in the original deleted row. The merge logic is here changed so that nulls in inserts are always interpreted as explicit null values.
- Simplified merge predicate: Previously we changed the merge predicate to handle null values in key columns so that these rows are applied correctly; however, this theoretically should never happen, and is less efficient to compute. This now uses a simpler predicate when there are no nulls detected in the column, and falls back to the previous behavior otherwise.
- A single duckdb connection is maintained for the whole job history; file footer information and so forth is therefore kept in cache and doesn't have to be re-read multiple times.
- `_check_load_records` now asserts that there are L records in a snapshot file, to avoid dropping data if we receive an invalid/empty snapshot file.
- `_check_load_records` and `_merge_apply` now assert that edw_inserted_dtm is non-null, to avoid the possibility of sorting rows into a broken month=0/day=0 partition.

### DIFF
--- a/src/odin/generate/cubic/delta_ods.py
+++ b/src/odin/generate/cubic/delta_ods.py
@@ -32,11 +32,13 @@ Two Qlik Replicate behaviors are relied on without a runtime check:
     ``ORDER BY header__change_seq LIMIT n`` and a strictly-greater watermark
     can never split records sharing a sequence across batches (which would
     permanently skip the cut-off records).
-  - "I" records are full row images (only "U" records may be sparse), so an
-    "I" is a per-key reset point: a winning "I" applies verbatim (NULL means
-    NULL), a winning "U" folds only across records at or after the key's
-    latest "I", and a batch containing an "I" replaces a matched target row
-    rather than coalescing with it. "D" images never contribute values.
+  - "I" records are full row images (only "U" records may be sparse). Both
+    "I" and "D" are therefore per-key *reset events*, and each key's latest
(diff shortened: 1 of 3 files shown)


src/odin/generate/cubic/delta_ods.py
```diff
@@ -451,22 +544,78 @@ def _dfm_candidates(self, cdc_df: pl.DataFrame) -> Iterator[str]:
             yield s3_file(path)
 
     def _build_merge_source(self, cdc_df: pl.DataFrame, keys: list[str]) -> pl.DataFrame:
-        """Resolve CDC records to one final row per key for the silver MERGE."""
+        """
+        Resolve CDC records to one final row per key for the silver MERGE.
+
+        Each key resolves to a single action, ``odin_resolved_oper``, decided by
+        the key's latest I or D record. Both are *reset events* — a delete ends
```
ealexa05, line 551: This is the correct way to handle this I think, I will note this assumption for future conversations with Cubic

ealexa05 (approved):
I think these changes are all correct and sensible. Will be good to monitor how duckdb performs as it approaches memory limit

PR #217, opened 2026-07-09
feat: Add disk monitoring to OdinJob

MaxAlex:
Disk usage from various processes spilling to disk is a common performance issue (particularly on AWS where the disk and CPU may not be proximate to each other) so this adds disk usage monitoring to the background OdinJob loop.

### DIFF
--- a/src/odin/job.py
+++ b/src/odin/job.py
@@ -223,9 +223,7 @@ def job_proc_schedule(job: OdinJob, schedule: sched.scheduler | None) -> None:
     return_manager = get_context("spawn").Manager()
     proc_return_val = return_manager.Value("i", NEXT_RUN_FAILED)
     proc_tmpdir_val = return_manager.Value("u", "")
-    proc = get_context("spawn").Process(
-        target=job.start, args=(proc_return_val, proc_tmpdir_val)
-    )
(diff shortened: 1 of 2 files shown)


src/odin/job.py
```diff
@@ -127,27 +135,59 @@ def _proc_tree_rss_mb(proc: psutil.Process) -> float:
     return rss / (1024 * 1024)
 
 
-def _monitor_proc_peak_rss_mb(proc: BaseProcess, poll_interval_secs: float) -> float:
+def _dir_size_bytes(path: str) -> int:
+    """Total on-disk size of all regular files under ``path``, in bytes."""
```
ealexa05, line 139: Do we have a sense of how long this takes? Or how much memory...

ealexa05: @MaxAlex

MaxAlex: It's just requesting file stats an accumulating into a single int, so there shouldn't be any particular memory usage. It could take a bit of time if there's a huge number of tmpdir files for some reason, but it's running in a separate thread from the main task in either case.

ealexa05 (approved):
Approving, I don't think this should take undue time to complete during any Odin job, as number of tmp files should always remain relatively low

PR #218, opened 2026-07-09
Add discovered key columns to log output

MaxAlex:
Adds the list of key columns to logging output from the delta fact table job.


ealexa05 (approved):
Thanks, this will be handy for debugging

PR #223, opened 2026-07-13
Changes for CubicODSDelta OOM errors

MaxAlex:
- Removes a redundant sort operation on the incoming CDC change table.
- Sets `SET preserve_insertion_order = false` since this was likewise redundant.
- Cap duckdb memory use at half of available RAM, since it can (hopefully) spill to disk whereas Python on Fargate can't.
- Removes `SET parquet_metadata_cache = true` to avoid invalid memory allocations from overlapping history table writes.

### DIFF
--- a/src/odin/generate/cubic/delta_ods.py
+++ b/src/odin/generate/cubic/delta_ods.py
@@ -55,6 +55,7 @@ from typing import Iterator
 
 import duckdb
 import polars as pl
+import psutil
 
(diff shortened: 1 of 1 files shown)


ealexa05 (approved):
All of these changes look sound. We should return to enforcing that a given table can never be updating history and fact table at the same time, but for regular history writes (i.e. not snapshot resets), this should work fine

PR #224, opened 2026-07-13
Remove several unnecessary log points

MaxAlex:
- Remove redundant CubicODSDelta top-level log, move fields unique to that to the top-level OdinJob logging.
- Remove logging from `pq_rows_and_bytes` and `pq_rows_per_mb`, since this was spamming the logs.
- `masabi_check_json_page` will now only emit logging if a schema discrepancy is found.

### DIFF
--- a/src/odin/job.py
+++ b/src/odin/job.py
@@ -86,12 +86,12 @@ class OdinJob(ABC):
         sigterm_check()
         self.reset_tmpdir()
         run_delay_secs = NEXT_RUN_DEFAULT
+        log = ProcessLog(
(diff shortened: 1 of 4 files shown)


src/odin/job.py
```diff
@@ -192,6 +192,7 @@ def job_proc_schedule(job: OdinJob, schedule: sched.scheduler | None) -> None:
 
     ProcessLog(
         "odin_job_peak_mem",
+        auto_start=False,
```
ealexa05, line 195: Good find on auto_start

ealexa05 (approved):

PR #225, opened 2026-07-14
Delta fact batch size 100k->200k

MaxAlex:
Increases the batch size, so we can observe impact on speed and memory usage.


ealexa05 (approved):
Approved. Take two, hopefully this time with more memory safety!

PR #226, opened 2026-07-14
Instantiate ProcessLog values in __init__() instead of start()

MaxAlex:
Recent update to logging using 'auto_start=False' has some odd results since start() actually instantiates some logging values, such as the start time. This moves those to __init__(). 'auto_start' isn't used outside said recent update, so this shouldn't affect anything else.


ealexa05 (approved):
Good catch, approved

PR #228, opened 2026-07-15
New mechanism for setting reschedule delay

MaxAlex:
OdinJob's reschedule themselves after each run; the desired delay can vary based on errors, status, or so forth. Most jobs have an internal notion of their intended reschedule times; meanwhile, OdinJob has its own default and on-error reschedule times, and all errors would result in the OdinJob default on-error reschedule time. This PR makes the 'run_delay_secs' an object property instead of a local variable, so the particular job subclass can set it as needed. This also includes a relevant change to the delta fact table job using this mechanism.

### DIFF
--- a/src/odin/generate/cubic/delta_ods.py
+++ b/src/odin/generate/cubic/delta_ods.py
@@ -54,6 +54,7 @@ import tempfile
 from typing import Iterator
 
 import duckdb
+from duckdb import OutOfMemoryException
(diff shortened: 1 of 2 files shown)


ealexa05 (approved):
Looks good to me; self.run_delay_secs isn't being set anywhere else in Odin that I can see, so this should be safe

PR #232, opened 2026-07-20
feat: adding status logs for each table to S3

ealexa05:
Users have no good way of determining when data is behind, often noticing only when reports become obviously incomplete. We have a few tools that we can use to dig in and identify problems when they arise (and when we are notified), but each data source works differently, the available metrics from each are different, and each involves some manual effort.

This PR introduces new, lightweight code which logs the status of each table to S3 as a JSON, which is easily inspectable by users (and us as maintainers), and structured for use in scripts. Specifically, it measures if and by how much the user-facing tables lag behind the history, API, and/or clock time, depending on what exact metrics available in that source. These logs exist for ODS tables (current-gen and deltalake based), S&B tables, and Masabi tables.

For example, S&B data tracks data by 'job_id', so a table which is behind might look like this (as measured on dev):

```
{
  "table": "v_shiftevent",
  "table_type": "transactional",
  "last_run": "2026-07-17T19:21:30.388014+00:00",
  "row_count": 3934207,
  "object_count": 1,
  "total_size_bytes": 24222679,
  "max_job_id": 559242,
  "api_latest_job_id": 569252,
  "jobs_lag": 9999,
  "rows_lag": 8261871,
  "lag_truncated": true,
  "next_run_seconds": 300,
  "run_duration_seconds": 25.29,
  "previous_last_run": "2026-07-17T18:02:02.627098+00:00",
  "previous_row_count": 3933049,
  "seconds_since_last_run": 4767,
  "rows_added": 1158,
  "rows_per_second": 45.79,
  "catchup_processing_seconds": 180429
}
```
Here we can see that `v_shiftevent` user-facing table, last run on `2026-07-17T19:21:30`, has `max_job_id` of 559242, which is behind the latest job_id available upstream in the API (`api_latest_job_id: 569252`). The `jobs_lag` measures how many jobs it is behind (capped at 9999, because it only fetched the latest 10000 jobs from the API, which is why `"lag_truncated": true`). These job_ids comprise 8261871 rows of data (`rows_lag`). In this run, `1158` rows were added, over 25.29 seconds, resulting in `rows_per_second": 45.79`, from which we can infer this table needs 180429 seconds of additional processing time (~50 hours).

ODS tables (of either type) track change data with `header__change_seq`, which contains the datetime that the change was generated by Qlik. This allows us to compare in terms of calendar time how behind each table is both in reference to our compiled history and to the upstream data generation. The ODS logs also contain information about the snapshot date and total rows, to assess how long it would take to catch back up if the table needed to be reset.

Masabi similarly records the date (in Unix time) that data was created on its server, which can be compared to the clock time.

The time added to each of these jobs to collect the data and write out the logs should be negligible.

### DIFF
--- a/src/odin/generate/cubic/delta_ods.py
+++ b/src/odin/generate/cubic/delta_ods.py
@@ -49,8 +49,10 @@ Steps per run:
 """
 
 import os
+import re
 import sched
 import tempfile
+import time
 from typing import Iterator
 
 import duckdb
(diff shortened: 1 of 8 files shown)


MaxAlex (approved):
All very lightweight (except for the delta_ods functions, which will be superceded by the delta history table change) for the functionality we get out of it; looks good.

PR #233, opened 2026-07-20
Retry faster on -11 flags (SIGSEGV errors)

MaxAlex:
Currently, all uncaught exceptions are retryed at OdinJob's default NEXT_RUN_FAILED delay, which is a full 24 hours. This makes sense when the issue e.g. vendor-side API issues, but less sense when the issue is sporadic runtime errors such as OOM hits. This catches the SIGSEGV error code and uses a shorter retry delay.


ealexa05 (approved):
Approved. This is a reasonable short-term fix for the intermittent SIGSEGV errors we are seeing due to ~~read/write conflicts caused by writing and reading history simultaneously, which should only be a factor during this deltalake transition~~ a duckdb version issue(?)

PR #235, opened 2026-07-20
Pin duckdb==1.5.4

MaxAlex:
Pins duckdb to version 1.5.4 (was previously 1.4.4 on prod) in case this fixes a series of invalid allocations and segfaults that appear to be coming from duckdb operations in delta_ods.


ealexa05 (approved):
We should start pinning more of these! On discussion, 1.5.4 seems like a reasonable target for now, though 1.4.x is the LTS version

PR #239, opened 2026-07-23
Add Masabi pytests

MaxAlex:
Following the recent incident where an update to logging caused Masabi updates to fail, I noticed that we don't currently have real CI tests for the Masabi job. As an immediate fix this PR adds an AI-generated test suite; I've confirmed that it would fail (a bunch) on the logging error, so similar issues will be mechanically prevented from merging in the future.


ealexa05 (approved):
I've checked through the tests and basic logic / test cases, they seem sound and I'll be glad to have the coverage!

PR #241, opened 2026-07-24
chore: adding delta instance

ealexa05:
Setting up a new instance (unfortunately named `delta`, but will likely be temporary) for syncing USE_TRANSACTION


MaxAlex (approved):
Approving, but "delta" can't be the name of an instance we keep indefinitely

PR #242, opened 2026-07-24
Add WSP611 and WSP620 views

MaxAlex:
Asana: https://app.asana.com/1/15492006741476/project/1214070495342435/task/1216796874357067?focus=true and https://app.asana.com/1/15492006741476/project/1214070495342435/task/1216176708986268?focus=true

Adds a new set of view definitions, by request. I have verified that these run against the data repo, although some are small; the WSP611 views are only a few dozen rows each.


ealexa05 (approved):
Looks good to me! Doesn't include any of our significantly behind tables, so the reports themselves must just be small? Can check if it is expected that the operation/system performance checks don't run very often, or started recently

PR #243, opened 2026-07-24
fix: fixing fact table partition-access OOM issue

ealexa05:
This PR addresses an OOM error repeatedly hit by CubicODSDelta while merging updates into the silver table. This root cause might be related to partitions being selected to match values of `odin_year` and `odin_month` appearing in the updates, which resulted in loading each possible month for each possible year (e.g., if the update df contained updates for (2024, 1) and (2025, 12), this would load (2024,1), (2024,12), (2025,1), and (2025,12)). The most recent batch included a handful of updates from prior years, which caused it to load many more partitions. The new code loads pairs explicitly.

### DIFF
--- a/src/odin/generate/cubic/delta_ods.py
+++ b/src/odin/generate/cubic/delta_ods.py
@@ -764,12 +764,9 @@ class CubicODSDelta(OdinJob):
         """
         Build the MERGE match predicate (keys + optional partition constraint).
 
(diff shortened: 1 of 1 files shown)


MaxAlex (approved):

MaxAlex:
These changes went in with https://github.com/mbta/odin/pull/248, so I'm closing this PR.

PR #248, opened 2026-07-27
Delta fact merge memory improvements

MaxAlex:
Multiple changes to the delta fact table job to minimize memory usage; this includes changes from Ted's PRs https://github.com/mbta/odin/pull/243 and https://github.com/mbta/odin/pull/247. 

In rough order of size/importance:

1. `_partition_constraint` is now constructed to select partitions by (odin_year, odin_month) pairs rather than matching either odin_year or odin_month, which was unnecessarily inclusive. This will directly reduce the number of fact table files touched.
2. Merging is now done in smaller batches by partition; each group by (odin_year, odin_month) is merged separately.
3. The duckdb connection is now closed once CDC batch has been read, since it may hold on to allocated memory even after the query is complete.
4. streamed_exec, which we recently set to True, is now False
5. merge() is now given settings for spilling to disk rather than simply running out of memory.

Notably, 2 requires two additional steps for correctness: 
- If a merge chunk is empty, merge() is a total no-op and doesn't apply metadata changes; this way, the watermark metadata could be lost. We therefore make it it's own separate update, in `_commit_watermark`
- edw_inserted_dtm must be immutable; if it's changed, the corresponding row would appear in multiple chunks, potentially leading to inconsistent results. We therefore assert this immutibility in `_check_partition_stability`


ealexa05 (approved):
Approving, new changes look good

PR #250, opened 2026-07-29
Remove max_spill_size setting, since this doesn't seem to help

MaxAlex:
Previous ods_delta PR included several changes, one of which was use of the `max_spill_size` and `max_temp_directory_size` parameters; under further inspection these turn out not to work as expected and have interactions with the instance core count that are difficult to calculate, so we're taking these back out.


ealexa05 (approved):
Approved, hopefully we can bump up the row size limit

PR #252, opened 2026-07-31
Add z-order sort for delta merge efficiency

MaxAlex:
It turns out that some tables, including SALE_TRANSACTION, don't usually supply edw_inserted_dtm on updates; since this is the source for the odin_year and odin_month columns we currently use for partitioning, the partitioning system doesn't work for these updates and the entire structure has to be scanned to locate the row to be updated. This is probably the source of a lot of the performance issues we've been seeing over the past week.

Several changes to mitigate this:

1. It now sorts the base-level files by the primary key, which is guaranteed to be present in all updates and which in most cases corresponds to the date. This is done using delta-rs' z_order() function, which sorts based on a set of interleaved key; on one key, as it's used here, this is equivalent to a standard sort. For the sake of performance, this is targeted only to partition directories which have been updated on any given run. (Worth noting that the preferable solution (what this is hackily using z_order to replicate) is deltalake's liquid clustering, which delta-rs does not currently support.)
2. We add a step to sort the changes by key value and bound the maximum size of a chunk; this means that both updates that don't include edw_inserted_dtm, and also large batches that all hit one partition, don't hit OOM-causing numbers of target files (instead partitions get incrementally updated by key value.)
3. Set the target_file_size parameter, which tells delta-rs to split files smaller; this allows tighter targeting of updates in key-order, so less data that has to be rewritten per update overall.

### DIFF
--- a/scripts/get_table_status.py
+++ b/scripts/get_table_status.py
@@ -117,9 +117,7 @@ def _fmt_count(value: Any) -> str:
     return f"{value:,}" if isinstance(value, int) else str(value)
 
 
-def fetch_group(
-    prefix: str, tmpdir: str, only: Optional[set[str]] = None
-) -> dict[str, dict]:
+def fetch_group(prefix: str, tmpdir: str, only: Optional[set[str]] = None) -> dict[str, dict]:
     """
     Download status objects under `prefix`; return {table: payload}
     """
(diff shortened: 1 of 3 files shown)


ealexa05 (approved):
Nice work! Approving to test on dev. Reminder to double-check the exact consequences of hitting an OOM error in the middle of the z-order, especially running against S3, and verify behavior if there are multiple key columns

PR #255, opened 2026-08-03
feat: better cubic table manifest+settings system

MaxAlex:
This PR changes how we track Cubic tables and their assignment to jobs/instances. We want to be able to set which job runs on which odin instance, on a table-by-table basis, without managing multiple large lists of tables (such adding a new table is a change in one place.) This replaces a patchwork system where e.g. history table jobs and (legacy) fact table jobs worked from the same table list, so one couldn't be disabled for a table without the other also being disabled.

For each table we now have a set of values, one per job type, which set the specified instance to run the job on that table. Per-instance lists of tables for each job are then compiled from this list.

Future table additions will simply add a line to TABLE_MANIFEST with the appropriate instance values. Likewise if we add new Cubic table handling jobs (e.g., delta history tables) this will be a fourth instance value 'column.'

### DIFF
--- a/src/odin/ingestion/qlik/tables.py
+++ b/src/odin/ingestion/qlik/tables.py
@@ -109,7 +109,7 @@ TABLE_MANIFEST = {
     "EDW.RIDER_CLASS_DIMENSION": ["alpha", "alpha", None],
     "EDW.RIDE_TYPE_DIMENSION": ["alpha", "alpha", None],
     "EDW.ROUTE_DIMENSION": ["alpha", "alpha", None],
-    "EDW.SALE_TRANSACTION": ["alpha", None, "gamma"],
+    "EDW.SALE_TRANSACTION": ["gamma", None, "gamma"],
     "EDW.SALES_CHANNEL_DIMENSION": ["alpha", "alpha", None],
     "EDW.SALES_SUMMARY_BY_DAY": ["alpha", "alpha", None],
     "EDW.SALE_TXN_PAYMENT": ["alpha", "alpha", None],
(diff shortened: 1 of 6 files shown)


src/odin/ingestion/qlik/tables.py
```diff
@@ -1,216 +1,162 @@
...
-    + CUBIC_ODS_DELTA_TABLES_GAMMA
-    + CUBIC_ODS_DELTA_TABLES_DELTA
-)
-
-_ODIN_INSTANCE = get_odin_instance()
-
-CUBIC_ODS_DELTA_TABLES_INSTANCE = CUBIC_ODS_DELTA_TABLES_BY_INSTANCE[_ODIN_INSTANCE]
+CUBIC_ODS_DELTA_TABLES = [
+    t for t, inst in TABLE_MANIFEST.items() if inst[DELTA_FACT_JOB_IND] == _ODIN_INSTANCE
+]
```
runkelcorey, line 163: what guards does this implementation provide against overwriting data? It looks like
1. it prevents more than 1 definition per key
2. it prevents the same table from being written to by multiple instances

Are there edge cases to consider here? For instance, what happens **during** deployment, as the ECS are being spun down and brought back?

MaxAlex: Each job per instance is interrupted when the ECS is taken down; jobs are (supposed to be) safe under shutdown signals, so this is fine. There's checks to ensure that multiple copies of the same instance don't run simultaneously (`check_for_parallel_tasks()` in utils/ecs.py) so we should never see the new deployment interfering with the old.

The only way we could lose data, in terms of scheduling, is by having two instances write to the destination simultaneously. That's prevented by this setup.

One thing that's _not_ handled is issues from history tables being written as the fact table job is reading them, which has been seen to cause (data-safe) errors. It occurs to me that we've stopped seeing that precisely because the history job wasn't running for any of the delta tables; I've moved those tables' history jobs over to the corresponding instance, which should also avoid that problem.

runkelcorey: Nice! I learned something today :)

runkelcorey (approved):

PR #256, opened 2026-08-04
Delta ods huge batch size

MaxAlex:
The previous changes introducing z_order sorting and merge batching should cap the amount of memory used in the merge process. This was the only point giving us OOM errors, so we should (in theory) be able to raise the load batch size much higher than before.


runkelcorey (approved):
Great! Let's see how it goes

PR #257, opened 2026-08-05
Add gamma instance history job, and others

MaxAlex:
Consolidating a few small changes into a single PR:

- Adds the ODS fact table job to odin-gamma in run.py, so that (along with the changes to table assignment from yesterday) history tables for SALE_TRANSACTION and UNSETTLED_CRDB_ACQ_CONF will update
- Finishes implementing the prevention fix for the Cubic regression issue; snapshot values are now checked against a regex to ensure they're something valid, and if this fails the job throws an exception.
- Cleanup table status script change

### DIFF
--- a/src/odin/generate/cubic/ods_fact.py
+++ b/src/odin/generate/cubic/ods_fact.py
@@ -169,7 +169,7 @@ class CubicODSFact(OdinJob):
 
         try:
             self.fact_snapshot = str(fast_last_mod_ds_max(self.s3_export, "odin_snapshot"))
-            if re.match('[0-9]{8}T[0-9]{6}Z', self.fact_snapshot) is None:
+            if re.match("[0-9]{8}T[0-9]{6}Z", self.fact_snapshot) is None:
                 # If fact_snapshot doesn't match expected format, fail to prevent data loss case
                 raise ValueError("Data exists but found invalid fact_snapshot.")
         except IndexError:
--- a/src/odin/generate/cubic/ods_fact.py
+++ b/src/odin/generate/cubic/ods_fact.py
@@ -1,10 +1,10 @@
 import os
 import sched
-import re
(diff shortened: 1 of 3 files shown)


src/odin/generate/cubic/ods_fact.py
```diff
@@ -168,8 +169,8 @@ def snapshot_check(self) -> None:
 
         try:
             self.fact_snapshot = str(fast_last_mod_ds_max(self.s3_export, "odin_snapshot"))
-            if self.fact_snapshot is None:
-                # Fail to prevent data loss case
+            if re.match("[0-9]{8}T[0-9]{6}Z", self.fact_snapshot) is None:
```
runkelcorey: Could you parse this as a time? Seems a bit safer

src/odin/run.py
```diff
@@ -73,8 +73,10 @@ def start():
 
     # Schedule ODIN Jobs
     schedule_sigterm_check(schedule)
-    if odin_instance in ["alpha", "beta"]:
+    if odin_instance in ["alpha", "beta", "gamma"]:
```
runkelcorey, line 76: would it help to make this more declarative? like some dict of with instances as keys and values as a list of functions?

MaxAlex: This system is going away altogether soon; most jobs (aside from DataDictionary) will be assigned just by setting table lists. That's waiting on changing around table settings for Masabi and AFC.

src/odin/generate/cubic/ods_fact.py
MaxAlex: Changed to parse with datetime; I assume this is what you meant? Is indeed more stringent, and will still throw a ValueError on an invalid string.

runkelcorey (approved):

PR #258, opened 2026-08-05
Allow exceptions for schema prescision

MaxAlex:
Asana ticket: https://app.asana.com/1/15492006741476/project/1208949713596457/task/1217168861303026?focus=true

History table jobs for EDW.UNSETTLED_CRDB_SYS_CONF and EDW.PAL_CONFIRMATION are currently stuck due to slight schema drift:

```
{'ordinal': 4,
  'name': 'PAL_CONFIRMATION_KEY',
  'type': 'NUMERIC',
  'length': 0,
  'precision': 10,
  'scale': 0,
  'primaryKeyPos': 1}
```
  
Becomes

```
{'ordinal': 4,
  'name': 'PAL_CONFIRMATION_KEY',
  'type': 'NUMERIC',
  'length': 0,
  'precision': 19,
  'scale': 0,
  'primaryKeyPos': 1}
```

In order to fix this without altering the (widely-used in other contexts) `ds_from_path` where this appears, we change the DFM load util to catch this case and substitute in the expected precision. This is emphatically a _temporary fix_ to get these tables running, with minimal risk or disruption, while we wait for a fix from Cubic.

### DIFF
--- a/src/odin/ingestion/qlik/dfm.py
+++ b/src/odin/ingestion/qlik/dfm.py
@@ -65,6 +65,42 @@ class DFMdataInfocolumns(TypedDict):
     primaryKeyPos: int
 
 
+class DFMprecisionException(TypedDict):
+    """Column matcher that pins a drifted DFM precision back to its archived value."""
+
+    name: str
+    type: str
+    scale: int
+    drift_precision: int
+    archive_precision: int
+
+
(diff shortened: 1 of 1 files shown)


runkelcorey (approved):
Has Cubic told us when we should expect a fix? If not, what do you think a permanent fix looks like?

PR #262, opened 2026-08-07
Update views to use delta versions of relevant tables

MaxAlex:
Updates all mentions of sale_transaction, use_transaction, edw_unsettled_crdb_acq_conf and cch_afc_transaction to use the delta table versions instead of the legacy fact table versions. (All views are referencing the catalog table views created in duck_db.py, so this is a drop-in replacement.)


runkelcorey (commented):
is the schema identical? or could there be partition-based filtering that is affected?

ealexa05 (approved):
Approved. I double-checked that the new schema is working in duckdb, and the schema is strictly additive as mentioned above, so this should just work

PR #263, opened 2026-08-10
Reenable DEVICE_LAST_STATE legacy fact job on odin-gamma

MaxAlex:
On Friday I moved DEVICE_LAST_STATE to delta processing on odin-delta; however, it turns out that this table is absolutely gigantic and is projected to take months of processing time to catch up even on delta tables. This PR reenables the legacy job, and moves all processing to gamma, which is currently less utilized (since USE_TRANSACTION is caught up.)


ealexa05 (approved):
Approved!

PR #266, opened 2026-08-11
More view rearrangement

MaxAlex:
Updates materialized views to use relevant delta tables. Also updates WO118 likewise, which I missed in the previous PR.


ealexa05 (approved):
Looks good; for good measure, I went through all the views to make sure you didn't miss any (materialized or non-materialized) views for edw_use_transaction or edw_sale_transaction

PR #267, opened 2026-08-11
Update CCH_AFC_TXN views to delta tables

MaxAlex:
Updates views to use CCH_AFC_TRASNSACTION delta tables, as those are now up-to-date.


ealexa05 (approved):
Approved; this appears to be all of the instances of `EDW_CCH_AFC_TRANSACTION` in the views

PR #272, opened 2026-08-13
Remove migration 0005 that Odin currently doesn't have permissions for

MaxAlex:
Takes out the migrations that would have removed old table status files, in case we want to push something to prod while we still don't have those permissions.


ealexa05 (approved):
We'll get those json files next time

PR #273, opened 2026-08-18
Add checkpointing to delta table update

MaxAlex:
We've been getting intermittent errors computing the edw_farerev_payg_trip_txn view (which is used in computing the larger COMP B views.) The error was clearly related to duckdb's delta table integration but was otherwise ambiguous:

`8754e949d08b ERROR uuid=ac804598-f161-48b5-9d0c-3dc98cd9af5f, _duckdb.IOException: IO Error: Failed to unpack scan data from kernel: DeltaKernel ObjectStoreError (8): Error interacting with object store: Generic S3 error: HTTP error: request or response body error`

Analysis from Claude suggests that this was due to the delta-kernel-rs behavior when reading delta tables which have multiple JSON commit files: it opens HTTP requests for all commit files since the latest checkpoint _simultaneously_, and then if duckdb can't work through the streaming data from all of these before S3's request timeout hits, this error occurs. The default behavior for delta-rs is to checkpoint after accumulating 100 JSON commit files, so this can be up to 100 simultaneous HTTP reads. (And the issue is probably exacerbated by the fact that we limit duckdb's to single-threaded operation here for memory reasons.)

Solution is to explicitly checkpoint after an update, so that we don't accumulate commit files at all, and delta-rs only has to read the (more efficiently handled) checkpoint parquet files for this step. This is a bookkeeping operation in the delta-rs interface, so shouldn't have any impact on data integrity.

### DIFF
--- a/src/odin/generate/cubic/delta_ods.py
+++ b/src/odin/generate/cubic/delta_ods.py
@@ -254,6 +254,7 @@ class CubicODSDelta(OdinJob):
 
             next_run = self._merge_cdc(cdc_watermark)
             self._write_status(next_run)
+            self._checkpoint_log()
 
             self.start_kwargs.update(
                 {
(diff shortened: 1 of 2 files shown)


MaxAlex:
@runkelcorey I think you asked to be tagged here. Good pitfall of delta tables to be aware of.

ealexa05 (approved):
Approved, interested to see how this affects run times

PR #281, opened 2026-08-25
Suppress missing cubic snapshot error

MaxAlex:
Tables with missing snapshots previously threw an exception which was logged as an error. This generated unnecessary error reports and would clutter up the Odin Sentry dashboard, so this PR now suppresses that exception (although occurrences are still logged.)


ealexa05 (approved):
Approved; noting that this in particular was coming to bite us when there was data in prod, but not in dev, and that we now have better alerting for stale tables

